PR #19, opened 2013-12-13
Add FITS_tools 

keflavich:
Sorry to be making 2 independent changes at once.... the web interface makes that easier.  If requested, I'll do the PR's properly.

EDIT: this just adds FITS_tools; re-titled it to avoid confusion

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -11,7 +11,7 @@
         {
             "name": "astroquery",
             "maintainer": "Adam Ginsburg <adam.g.ginsburg@gmail.com>",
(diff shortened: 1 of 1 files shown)


eteq:
@keflavich - I was just going through some of the old issues here - is this still relevant?  That is, do you still want this to be considered?  If so, can you post on astropy-dev for discussion?

keflavich:
FITS_tools is still relevant, but since I made this PR, two new affiliated
packages that include a large fraction of its functionality have shown up:
imageutils and reproject.  So I guess we may as well close this.

PR #44, opened 2014-11-27
Fixes #42 seperates the table based on provisional column

### LINKED ISSUE #42: Separate table of provisional vs accepted packages

sudheesh001:
screen shot 2014-11-27 at 3 26 26 pm

### DIFF
--- a/affiliated/index.html
+++ b/affiliated/index.html
@@ -122,10 +122,10 @@
 	</section>
 	<section id="affiliated-package-registry">
 
-	<h3>Affiliated Packages Registry</h3>
+	<h1>Affiliated Packages Registry</h1>
 
 	<p>The following table lists all currently registered affiliated packages. This table is determined from the <a href="http://www.astropy.org/affiliated/registry.json">json file</a>, which is the actual authoritative registry.</p>
-	<h2>Affiliated packages</h2>
+	<h3><u>Provisionally accepted Affiliated packages</u></h3>
 	<table border="1" class="docutils">
 	<colgroup>
 	<col width="17%" />
@@ -159,7 +159,7 @@
 	<p>The Stable column indicates whether the package maintainer consider the package to be ready for use. Packages that are under heavy development and for which the user interface is likely to change significantly in the near future are marked as No.</p>
 	<hr>
 	<!-- Not provisional -->
-	<h3>Provisionally accepted Affiliated Packages</h3>
+	<h3><u>Affiliated Packages</u></h3>
 	<table border="1" class="docutils">
 	<colgroup>
 	<col width="17%" />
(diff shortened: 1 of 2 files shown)


sudheesh001:
@astrofrog @eteq Could you have a look at this if possible ? I am looking forward to contribute more to this project and the website.

eteq:
@sudheesh001 - sorry about this, somehow the notification for this dropped off my radar and only just noticed it while looking.  I will aim to review this tomorrow.

sudheesh001:
Cool!! Looking forward to having this patch merged today.

eteq:
@sudheesh001 - a couple things:

1. It looks to me like you've got which is in which table backwards.  Packages where 'provisional' is 'false' are *not* provisional.  If they have a date, they *are* provisional.  But in the version you have here, that's swapped.  So I think you just need to switch them around.
2. Now that there's a separate section, it might be good to add a 'next review date' column to the provisional packages.  That's what the meaning of the dates are (when 'provisional' is not 'false').  If you'd rather not do that in this PR that's ok too, but it might make sense while you're thinking about it.

One other minor item: I see above that you've been squashing/amending your changes so that there's only one commit for this PR.  In general, we prefer that you *not* do this: it's better to actually keep a record of the changes, as well as the context of the comment thread here.  Squashing or amending should really only be done in special circumstances (i.e., you accidentally committed a password, or a large binary file that would make the repository awkwardly large).  So in the future I suggest instead you just push up new commits.

eteq:
@sudheesh001 - can you try to make the changes I've mentioned above?

sudheesh001:
@eteq , Fixed those changes of getting the changes on the table backwards. It looks like it's good to go now.

I think its best to file a new issue for the adding a next review date column for provisional packages and then i'd be more than happy to send a patch to that, once this is merged. I'll be opening an issue for this.


>One other minor item: I see above that you've been squashing/amending your changes so that there's only one commit for this PR. In general, we prefer that you not do this: it's better to actually keep a record of the changes, as well as the context of the comment thread here. Squashing or amending should really only be done in special circumstances (i.e., you accidentally committed a password, or a large binary file that would make the repository awkwardly large). So in the future I suggest instead you just push up new commits.

The other open source organizations i've pushed patches to always asked for this, so I guess it kind of became a habit :yum: Pushed the commit normally this time.

embray:
Hmm, not to cause confusion (sorry @sudheesh001) but I actually prefer that PR authors squash their commits, at least at the end of review / before merge.  I'm not saying it should be squashed down to a *single* commit, but typo fixes and other fixes made in response to review I think should be squashed down into a minimal, cohesive set of changes.  I do this myself manually on some PRs that have a mess of a history, though it's nice when contributors can do this themselves as well :)

Again, my preference is to do it just prior to merging though.

sudheesh001:
@embray Thats a good thing to follow, I'll squash it into one if the review passes from any of the Collaborators or Owners.

Issue #46, opened 2014-12-07
Add WCSAxes to affiliated package repo


astrofrog:
If we go ahead and merge https://github.com/astropy/astropy/pull/3183 then we should add WCSAxes to the list of affiliated packages (and I will move the repo to the astropy organization).

sudheesh001:
@astrofrog It looks like astropy/astropy#3183 has been merged. So this issue could be taken up and patched. Can you review if the following information is correct (and provide information where I missed out), I would be more than happy to send a Pull request.

```
Package Name: WCSAxes
Stable: 
Provisional: 
PyPI name: wcsaxes
Web page: https://github.com/astrofrog/wcsaxes
Code repository: https://github.com/astrofrog/wcsaxes.git
Maintainer: Thomas Robitaille
```

astrofrog:
@sudheesh001 - yes, this looks correct, thanks! Please set stable to Yes and Provisional to Yes.

sudheesh001:
Fixed in commit b05d746 This is good to close @astrofrog

astrofrog:
Thanks @sudheesh001!

PR #53, opened 2015-02-21
Fixes Issue #46 Adds WCSAxes to the affiliated package repo

sudheesh001:


astrofrog:
Thanks @sudheesh001!

eteq:
@sudheesh001 @astrofrog - one thing I noticed here just now: the "provisional" entry should *not* be "true", but rather it should be set to a date 1 year from when the package was first accepted as provisional.  I'll just fix this in master right now, but I wanted to mention this so we can avoid it in future PRs.

One other tip, @sudheesh001 : if I'm reading this right, merging didn't automatically close this issue. Text like ``"closes #xxx"`` in the commit log or issue description does this automatically.  I see you do have ``"closes issue #53"`` in the log, but it didn't work.  I think you have to instead use ``"closes #53"`` or ``"fixes #53"`` or similar (i.e., no "issue" part).  Just FYI - it makes it easier to keep track of all these issues.

(Same comment applies for #54)

PR #60, opened 2015-04-22
Add gwcs to affiliated package registry

cdeil:
This PR adds `gwcs` to the affiliated package registry to make it easier to find.

@eteq @astrofrog – OK?

@nden – You should register the name `gwcs` on PyPI to make sure it's still available when you want to make a release.


astrofrog:
:+1: from me.

eteq:
Looks good, will merge now.  @nden - I registered it on pypi now just so the info here is accurate.  If you can let me know your username on pypi I can change it so that you're an owner.

PR #67, opened 2015-04-27
Added glue as an affiliated package

astrofrog:
Based on an email discussion a little while back with @ChrisBeaumont and @perrygreenfield, I think we agreed that glue should be an affiliated package. I think this is a good time to make sure we also understand what the implications of being an affiliated package are. So just to be clear, listing glue as an affiliated package does not exclude having it also be affiliated with other projects (e.g. outside astronomy), correct? In my view, the key sentence is the following (from our website):

> ... but has requested to be included as part of the Astropy project’s community. **These packages are expressing an interest in Astropy’s goals of improving reuse, interoperability, and interface standards for python astronomy and astrophysics packages.**

Because glue is also intended for use outside astronomy, so we need to be sure that being affiliated does not mean that glue is 'part' of the Astropy project as such. I guess we could try and make these kinds of points clearer on the website to avoid discouraging people from becoming 'affiliated'?

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -127,6 +127,15 @@
             "pypi_name": "wcsaxes"
         },
         {
+            "name": "Glue",
+            "maintainer": "Chris Beaumont and Thomas Robitaille <thomas.robitaille@gmail.com>",
+            "provisional": "2016-04-27",
+            "stable": true,
+            "home_url": "https://www.glue-viz.org",
+            "repo_url": "https://github.com/glue-viz/glue.git",
+            "pypi_name": "glueviz"
+        },
+        {
             "name": "pyregion",
             "maintainer": "Jae-Joon Lee",
             "provisional": "2016-02-22",


wkerzendorf:
@astrofrog TARDIS also has the goal of reuse, interoperability and interface standards. Would you consider us?

astrofrog:
Of course! :)

ChrisBeaumont:
+1 from me :) Nothing about the current wording makes me too nervous that becoming an affiliated package would somehow corner a project as solely-astronomical.

perrygreenfield:
Likewise, I don't think being affiliated means it can't be used in other contexts. It just says that the package adheres to the principle of being compatible with astropy, it's documentation, and tests (though tests with GUIs are a very difficult issue).

cdeil:
For the end-user, probably the most important aspect of http://www.astropy.org/affiliated/ is a link collection, i.e. to find useful packages. Since it's not possible to list everything there (e.g. healpy, sherpa, ...), maybe a link to http://www.astropython.org/resources would be a good idea?

astrofrog:
@cdeil - yes, once astropython.org has been revamped (currently the astropython site contains packages we shouldn't really be recommending anymore and the plan for the new astropython site will be to make that kind of distinction clearer)

cc @taldcroft

astrofrog:
@eteq - I think this should be good to go

eteq:
@astrofrog @perrygreenfield - I'm fine with glue as an affiliated package, but I think perhaps for a larger one like this, which had an independent existence prior to being an affiliated package, a request should be posted to astropy-dev first, as is the at least nominally official procedure (and given a few days for any comments before we merge).  I don't expect anyone will object, but I'm a bit uncomfortable "short-circuiting" the procedure we've at least said we're following on the web site.

astrofrog:
@eteq - I can still send an email to the mailing list about this if you would be more comfortable. I had forgotten that emailing the list was part of the formal process.

eteq:
@astrofrog - yeah, I had too, but then I looked at the web site and that is what we said ;) 

I'll be making a PR shortly to update those instructions with the provisional review procedure, so we can re-consider the matter there.

eteq:
The PR I mentioned above is #71 - I suggest moving discussion of whether astropy-dev posting should be mandatory be moved there.

(And @wkerzendorf, absolutely feel free to ask for affiliated status for TARDIS whenever you feel it's appropriate.  Although as per the discussion above, definitely post to astropy-dev in parallel to a PR.)

eteq:
Alright, #71 is now merged - so I'd say at least briefly mention this on astropy-dev and give it a few days. I'm pretty sure no one will have any objections, but at the very least this means people will see it.

eteq:
@astrofrog @ChrisBeaumont - just as a reminder, the judgement in #71 is that you should put the request into astropy-dev so that anyone there can comment.

astrofrog:
@eteq - requested on https://groups.google.com/forum/#!topic/astropy-dev/B05JF9KomiI

eteq:
@astrofrog - Awesome.  If there are no significant objections by the end of the week, I think we can go ahead and merge this (the coordinating committee already discussed it, so no need to re-visit if there's not further input).

astrofrog:
@eteq - can we merge this?

eteq:
yes!

PR #58, opened 2015-03-26
HiDPI support -- use SVGs and fall back to PNGs

mdboom:

PR #69, opened 2015-05-05
Fix astropy logo so that index page matches other pages

eteq:
In master right now (and hence on the astropy.org site), the front page and all the other pages have slightly different sized "astropy" text (in the upper-left corner). This PR adjusts it so that they're all the same.

@mdboom - the git history says you set this up as part of #58 - was there some reason this was supposed to be different, or was this just an oversight?


mdboom:
Just an oversight.

embray:
Might have been partly me too, since I added this in several other pages after @mdboom's original change.  I might have proliferated one particular size, depending on what I copy/pasted from :)

eteq:
Alright, just wanted to make sure.  Merging now, then.

PR #70, opened 2015-05-05
Added Travis to validate JSON file

astrofrog:
Also, everyone loves it when Travis says it's ok to merge. Even if the changes are unrelated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+
+python:
+  - 3.4
+
+script:
+  # Validate JSON for affilaited package registry
+  - python -c 'import json; json.load(open("affiliated/registry.json"))'


embray:
While you're at it, why not add a JSON Schema for the file and validate it against the schema :D

embray:
Or better yet, a YAML Schema: https://github.com/spacetelescope/yaml-schema-standard

(unfortunately the code for that is still buried in pyasdf, but I still plan to separate it out for general use)

eteq:
@astrofrog - I think @embray might have been at least partly joking, but a schema validator isn't a bad idea if it's easy.  Do you know of an easy-to-use JSON validator? (Maybe https://github.com/mafintosh/is-my-json-valid although it's js-based)

I'm going to merge this now though, so that we have the "is it valid JSON" in now, but a follow on PR can always add the validator.

embray:
I was joking a little, but only like 20%, to the extent that for this particular purpose it's definitely overkill. But still good practice.  

For ASDF we use the [jsonschema](https://pypi.python.org/pypi/jsonschema) module which IMO is quite good, and @mdboom and I have worked on a few extensions to it that are also nice though maybe not necessary for this purpose.

astrofrog:
I won't have time to work on a json schema for this, and while it would be nice, I classify this under 'low-priority' :)

PR #71, opened 2015-05-05
Updated info on the provisional review process

eteq:
Prompted by @weaverba137's comment on astropy-dev (and after some discussion with @astrofrog and @perrygreenfield), I've put in a bit more about what the provisional affiliated package review process involves.  Hopefully this will clear stuff up and allow the coordination committee to get to actually doing these reviews.


eteq:
An item to consider in parallel with this: should we allow affiliated package requests via Pull Request and *not* via astropy-dev announcement?  We've been doing this some, but so far I think only only for "in-house" packages, i.e. those developed by people who are already part of the core dev team (i.e., @mdboom @embray @astrofrog).  Should we allow this for everything?  Or restrict it to nothing? 

My feeling is that there should be at least an astropy-dev posting for everything, going forward, just in case someone on the list who doesn't monitor this repo (i.e., almost everyone...) has objections/wants to be involved in the process of something they previously didn't know about.

astrofrog:
I think it makes sense to post to astropy-dev - if anything, it's good advertising.

I would suggest that we allow people to request a review before the 1 year period. If a package is ready, a year seems like a long time to wait.

eteq:
Both good points @astrofrog, I agree.  Updated to reflect the latter (the former item is already in there).

eteq:
(the Travis failure isn't real, it's an artifact of turning travis on for #70 )

eteq:
Alright, I'm going to go ahead and merge this, as the coordination committee has discussed this already offline.

PR #72, opened 2015-05-07
affiliated: add a link

keflavich:
link to bottom of page when it is mentioned


eteq:
Good call, thanks @keflavich

PR #75, opened 2015-05-14
add code of conduct

eteq:
If/when astropy/astropy-APEs#11 is adopted, this PR will add the code of conduct to the web site.


astrofrog:
I wonder if it should be a new top-level item, right of "Affiliated packages", to make it more prominent? It doesn't really fit in 'about'.

astrofrog:
Even if it's not its own heading, it should be its own page that we can link to from other places.

eteq:
@astrofrog - I guess I don't feel it deserves any more prominence than, say, the acknowledgements or authorship list, but I think promoting *all* of those things to the banner makes it too busy.  So I was thinking this was the safest way to go.

In the current PR it *is* linkable directly as  http://www.astropy.org/about.html#code-of-conduct  - but I don't have a strong opinion on that, so if you still think it should live on its own page, that's easy enough to do.

astrofrog:
I agree, let's just merge this for now - however, I wonder if we might consider something other than 'About' to describe this, like 'Community'. Will open another issue to discuss this.

PR #79, opened 2015-06-14
remove Other Packages section

eteq:
This makes a change to the affiliated packages site to remove the "Other packages" section.  It's been out-of-date for a while... I think the intent was to be text descriptions of the packages that are not "featured", but the number of packages has grown enough that that has become impractical (or at least it's not getting done).  So this just removes it.

To replace this, it might make sense to add a ``description`` section to the JSON registry and the package table. (I think @embray suggested this at some point?)  But until then I think it's better to remove this rather than leave it just very incomplete (and suggesting some of the packages are more special than others).

cc @astrofrog

### DIFF
--- a/affiliated/index.html
+++ b/affiliated/index.html
@@ -96,31 +96,6 @@
 		</div>
 
 	</section>
-	<section>
-
-	<h2>Other packages</h2>
-	<h3>Stable packages</h3>
-
-	<ul>
-		<li><a href="http://astroml.github.com/" target="_blank">astroML</a>: tools for machine learning and data mining in Astronomy</li>
-		<li><a href="http://pythonhosted.org/Astropysics/" target="_blank">Astropysics</a>: a library of assorted optical astronomy and astrophysics tools (in the process of getting ported to Astropy).</li>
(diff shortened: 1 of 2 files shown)


embray:
:+1:, also to filling in a description section to the JSON registry.  Authors of those packages should provide such a description if they want (or it can be taken from their package metadata if nothing else :)

eteq:
I updated this so that a "description" field exists for everything that was in Other Packages. Ready to merge if @astrofrog thinks this is ok.

Once #82 is in, I'll see about trying to add these descriptions back in in an automated fashion.  One problem is that the descriptions might make the affiliated package table awkwardly long.  A possible solution is to resurrect this "Other Packages" section but have it be automatically populated with the descriptions.

astrofrog:
@eteq - this looks good. I wonder whether we might make some of the other columns more concise, for example by simply saying 'GitHub' and having it link to the GitHub repository, rather than writing out the full URLs in the table? (same for Home page)

eteq:
@astrofrog - good idea, I'll create an issue for these things

PR #81, opened 2015-06-15
Use HTTPS for external JS and fonts.

### LINKED ISSUE #74: https://www.astropy.org/affiliated/ does not work correctly
Hi,

the SSL-enabled web page https://www.astropy.org/affiliated/ does not show the package registry. The table stays in the state "Loading...".

mdboom:
That way viewing the website through https will work.

Fix #74.

### DIFF
--- a/about.html
+++ b/about.html
@@ -6,7 +6,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1" />
 <link rel="shortcut icon" href="favicon.ico" />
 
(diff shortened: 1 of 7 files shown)


mdboom:
Actually, it turns out this breaks browsing files on the local file system.  I'll just make these all https: which should resolve it for all use cases.

embray:
I was just gonna say the same thing--might as well just specify https, which should definitely be used for loading external JS regardless if the originating site is using http.

eteq:
Looks good - sorry this has languished but I'll go ahead and merge now.

PR #82, opened 2015-06-20
clean up affiliated table js code

### LINKED ISSUE #80: Provisional/non-provisional packages split the wrong way around
@eteq - FYI the tables for provisional vs accepted need to be swapped. I can try and open a PR later.

eteq:
This fixes #80 and also generally makes the table-populating stuff more robust.  The result can be seen at http://eteq.github.io/astropy.github.com/affiliated/ (it's visually the same as the current version, but with affiliated and provisional swapped, as they should be to fix #80)

### DIFF
--- a/affiliated/index.html
+++ b/affiliated/index.html
@@ -129,7 +129,7 @@
 	
 	<!-- Not provisional -->
 	<h3><u>Affiliated Packages</u></h3>
-	<table border="1" class="docutils">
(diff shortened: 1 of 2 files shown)


eteq:
@astrofrog, do you want to review this?  I'll give it a day or two and if no one's gotten to it by then I'll just merge so that we get the #80 fix in soon.

PR #89, opened 2015-08-25
Adding package: astroplan

bmorris3:
Adding the [astroplan](https://astroplan.readthedocs.org/en/latest/) package. (cc @eteq )


cdeil:
Best package ever!  :+1:

eteq:
The coordination committee has agreed to accept this package! I rebased this and then merged in ff0de26eb24b0e6b81e8ff65857441596c2de9a1 - I think because of the rebase git doesn't know that's the same as this PR, but I promise it's in, so I will close this now.

cdeil:
FYI: I've made a package entry on astropython.org ... it needs moderator approval, but it should appear here: www.astropython.org/packages/astroplan

PR #90, opened 2015-09-11
Add naima to affiliated registry

zblz:
naima is a pure-Python package that allows to infer the properties of relativistic particle energy distributions from observed nonthermal spectra. It has two major components: a set of nonthermal radiative models (synchrotron, inverse Compton, bremsstrahlung, neutral pion decay), and a set of functions that use emcee to infer the particle distribution properties by fitting observed spectra through an MCMC procedure.

 - Documentation: http://naima.readthedocs.org
 - Code: http://github.com/zblz/naima
 - PyPI: https://pypi.python.org/pypi/naima


eteq:
As just mentioned on-list, this package has received full approval.  Congrats @zblz !

PR #95, opened 2016-01-26
Add section on installing affiliated packages

eteq:
This adds a section to the affiliated package page with installations tips.  Obviously we can't give a definite "this will always work" rule, as they can always be different.  But these instruction should work for the vast majority of cases.

Note that the first suggestion is "use the conda channel".  @mwcraig was said he wants to do some additional cleanup before we do an advertising push on that, so *THIS SHOULD NOT BE MERGED* until he gives the thumbs up.

There's also a plan in the works to re-organize the repo where the conda channel is stored, so the link to that repo in this PR needs to be updated when that happens (the "conda-builder-affiliated" will likely change) 

cc @astrofrog


mwcraig:
@eteq -- The reorganization is far enough along to merge this.

A couple changes:

+ New maintenance repo for the affiliated package channel is: https://github.com/astropy/conda-channel-astropy
+ The conda channel itself is at: https://anaconda.org/astropy/packages

Might be helpful to include a link to the latter for a full list of what is available in the channel.

eteq:
Alright, thanks @mwcraig - I updated this and will merge it now - feel free to issue another PR if you think something should be clarified/changed.

PR #97, opened 2016-03-09
Updating links to point to the webpage or docs page

bsipocz:
It makes no sense to list the link of the repos twice, so I've updated the ``home_url``s to point either to the webpage (if there is one), or to the docs page. In the odd cases there isn't any docs page yet, so then repo is left in place.


eteq:
Good point on all of this, @bsipocz - I'll just go ahead and merge this.

PR #104, opened 2016-03-30
add mention of gitter in the getting help page

eteq:
This updates the "getting help" page to point people to the astropy gitter channel. Prompted by @cdeil's suggestion on astropy-dev.

Note that, at least for now, I am *not* removing the mention of the IRC channel as it still has occasional people there.  But it's at the bottom of the list so we can probably phase it out if gitter does indeed catch on this time. 

cc @cdeil @astrofrog

### DIFF
--- a/help.html
+++ b/help.html
@@ -59,10 +59,9 @@
 	<p>The best way to get help is usually by asking questions of the Astropy user and development community. </p>
 
 	<ul>
-		<li>If you want to have a quick question and want immediate feedback, you can join <a href="https://gitter.im/astropy/astropy">Astropy's gitter channel</a>. Gitter is basically a live web chat, but with features that make it better for discussing code. </li>
+		<li>If you want to have a quick question and want immediate feedback, you can join <a href="https://gitter.im/astropy/astropy">Astropy's gitter channel</a>. Gitter is basically a live web chat, but with features that make it better for discussing code. If you prefer using IRC clients for chat, there is <a href="https://irc.gitter.im/">a gitter to IRC bridge</a>.</li>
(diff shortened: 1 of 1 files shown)


cdeil:
> This updates the "getting help" page to point people to the astropy gitter channel.

Thanks!

> Note that, at least for now, I am not removing the mention of the IRC channel as it still has occasional people there. But it's at the bottom of the list so we can probably phase it out if gitter does indeed catch on this time.

I don't see how keeping the IRC channel is helpful.
If you do ... is there a way to auto-post a reply "please go to Gitter" with a link on IRC?

eteq:
> I don't see how keeping the IRC channel is helpful.

One main reason I was thinking: People might balk at having to sign up for gitter (I don't think it's a big deal, myself, but people often complain about these things...).

But maybe you're right that we shouldn't advertise it on that basis?


> If you do ... is there a way to auto-post a reply "please go to Gitter" with a link on IRC?

That's a good idea regardless (as some people might just know the IRC approach already and not be paying attention to the web site).  I know there's lots of IRC bots out there, the main challenge is finding a place to host it.  Will look into it.

eteq:
Alright, the combination of @cdeil and @saim's n comment (which were on gitter, in a very meta discussion ;) convinced me to drop the IRC bullet point.  I did add mention of the gitter<->IRC bridge, though.

cdeil:
> People might balk at having to sign up for gitter (I don't think it's a big deal, myself, but people often complain about these things...).

On the other hand ... getting more astronomers to sign up on Github (which will give them access to Gitter) is a good thing and just takes a minute. :-)

eteq:
Alright, we updated the readme in astropy/astropy#3695, so now seems a good time to merge this

PR #110, opened 2016-06-25
add numfocus to about page

eteq:
I just realized we never actually got to adding the fact that numfocus fiscally sponsors Astropy to the about page.  This PR rectifies that

cc @astrofrog


astrofrog:
:+1:

PR #113, opened 2016-08-05
Search for issues filed only under org `astropy`

crodjer:
Right now the links on http://www.astropy.org/contribute.html point to all of github. This patch changes them for to only projects under [astropy](https://github.com/astropy) repositories.


bsipocz:
@crodjer - I think it's fine to leave as it was to be able to pick up the issues from more affiliated packages. Browsing through a few pages of the ``package-novice`` and ``effort-low`` issues, I could only see astropy and affiliated packages (and sunpy).

crodjer:
Thanks for the clarification, closing.

PR #121, opened 2016-09-01
Change contact email to coordinators@astropy.org

taldcroft:


taldcroft:
I need to change the contact link in `update_roles.py` as well.  Will do this later today.

PR #125, opened 2016-09-27
Add Miguel de Vol-Borro in Arch linux role

taldcroft:
> I will be glad to have a lead or deputy role for Arch Linux packaging.
> 
> Best,
> Miguel


saimn:
I think @mboquien should also be in the loop, as the current maintainer:
https://aur.archlinux.org/packages/python-astropy/

PR #128, opened 2016-09-27
Migrate roles update.py to Py3 (req'd for unicode names)

taldcroft:
Adding "Sébastien Fabbro" to the team requires Python 3.


bsipocz:
Oh, this case I can now have my accent, too :) (though maybe it's better to stay without it)

taldcroft:
PR's accepted!  :smile:

PR #129, opened 2016-09-27
Add Sébastien Fabbro as Gentoo packager

taldcroft:
Sébastien Fabbro has volunteered and committed to this role for at least a year.


granttremblay:
Looks great, thanks Sébastien!

PR #136, opened 2016-10-26
Update lead and deputy for astropy.constants

taldcroft:
This adds Tom Aldcroft as "process-only" lead and David Shupe as deputy with primary responsibility for technical review and development.


eteq:
@taldcroft - I wonder if the "Responsible primarily for process" footnote should be combined with "Can take lead while training new  deputy / future lead"?  I don't think we want to have too fine-grained notes like this, as it leads to confusion over what the purpose of the roles is...

eteq:
Looks good.  Merging!

PR #146, opened 2017-01-23
Add @bsipocz as releaser

eteq:
Adds @bsipocz to the roles as a deputy release manager.

Also removes the footnote for @astrofrog (is that ok?  I can back out the commit if not, @astrofrog)


eteq:
(got a 👍  from @astrofrog out-of-band)

PR #162, opened 2017-06-09
make whats new centered and include current version

eteq:
This is one of the things I said I'd like to modify in #159 (https://github.com/astropy/astropy.github.com/pull/159#issuecomment-307214063).  It updates the front page to include the current version in the "what's new" section, looking something like this:

image

I think this also actually addresses my "wall of text" concern in #159 now that I see it.  What do you think about this, @abostroem and @kelle?

PR #164, opened 2017-06-09
Make sidebar be the gradient instead of pure orange

eteq:
While working on #162 and #163 I noticed that when the sidebar appears with the menu (more importantly, it's what you see on a phone), you get this color:
image

That solid orange was very glaring and painful (at least on my monitor), so this PR changes it to match the gradient that the "buttons" on the rest of the page look like:
image

I think this is much more pleasant to look at.  This is quite subjective though, so it would be good to hear if others agree!

### DIFF
--- a/css/jquery.sidr.light.css
+++ b/css/jquery.sidr.light.css
@@ -15,6 +15,19 @@
 	-o-transition: none;
 	-ms-transition: none
 	transition: none;
+	background-image: -webkit-gradient(
+	linear,
+	left top,
+	left bottom,
+	color-stop(0, #FC9468),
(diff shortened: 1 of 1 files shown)


eteq:
Also, note that you can look at this "live" at http://eteq.github.io/astropy.github.com/ (although it's mixed with #162 and #163)

eteq:
@granttremblay - just FYI, in general I try to not merge my *own* PR's as a way to not abuse my powers any more then necessary 😉 ... So if you think this is good I'll leave it to you to merge.

granttremblay:
Ha! okay, good strat. Merging.

PR #165, opened 2017-06-22
A little cleanup of affiliated package json

mwcraig:
This should be looked at by @astrofrog because of the second commit, which removes WCSAxes. 

That may not be the way to handle packages that have migrated into core, but I was looking at the json anyway and figured I'd raise the issue. Happy to revert if desired....

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -110,7 +110,7 @@
         },
         {
             "name": "ccdproc",
-            "maintainer": "Steven Crawford and Matt Craig <ccdproc@gmail.com>",
+            "maintainer": "Steven Crawford, Matt Craig and Michael Seifert <ccdproc@gmail.com>",
             "provisional": false,
             "stable": true,
             "home_url": "http://ccdproc.readthedocs.io",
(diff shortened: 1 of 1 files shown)


bsipocz:
Maybe you could add ``regions`` to the list here? We do point people to that package even though the long term plan is to get it into core in the future.

astrofrog:
:+1: to removing WCSAxes from the list to avoid confusion

@bsipocz - yes, I think that regions should probably be affiliated - after all it's not clear how soon it will actually be merged into the core, and we should give it some visibility for testing in the mean time (@eteq: for in-house packages like this, do we skip the normal review process and mark it as unstable or provisional?)

eteq:
@astrofrog - That is indeed what we've done in the past.  We've never set up anything formal about that... But that's exactly what we did for `specutils` and `photutil` originally.  So I think that's a fair precedence here. Do you think we should wait for the upcoming changes to the affiliated package machinery, though?

bsipocz:
If this is held up only because regions, could we just merge now as is and deal with that separately?

eteq:
lets just merge this to clean things up.

PR #169, opened 2017-07-07
added release announcement for 2.0

eteq:
This PR adds the release announcement for v2.0.

One semi-questionable item here: I've included the mention that this is the last 2.x-compatible version in the list of "New and improved major functionality".  It's not clear that it's literally that, but it seemed one of the best way to get people's attention - it's also mentioned in the bottom paragraph, but that's not as obvious as it seems like that should be.  But it could be put above, as well, if even more prominence is desired.

You can see this rendered at http://eteq.github.io/astropy.github.com/announcements/release-2.0.html


astrofrog:
I made a few fixes - the biggest one is that 3.0 is *not* an LTS release (see the table in https://github.com/astropy/astropy-APEs/blob/master/APE10.rst). There's no reason to support two LTS branches at the same time.

bsipocz:
I've updated the version number as part of the PR, hope it's OK.

PR #173, opened 2017-07-11
Remove WCSAxes

olebole:
WCSAxes is marked as deprecated on [its github repository](https://github.com/astrofrog/wcsaxes), and superceded by `astropy.visualization.wcsaxes`.
To avoid confusion, I would propose to remove it from the affiliated package list as well.

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -128,15 +128,6 @@
             "pypi_name": "sncosmo"
         },
         {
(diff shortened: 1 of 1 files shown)


bsipocz:
See #165

olebole:
Ah, OK; closing here then. I would just now wait for tool long for the obvious change...

PR #175, opened 2017-07-25
Add Inclusivity Statement to all page footers

granttremblay:
Per issue #145, here is a very simple implementation of a unified Inclusivity Statement, added to the footer of every page (including the landing page). The link brings the user to the Code of Conduct. A small logo reinforces branding. Nothing else added to keep things clean. 

Full implementation can be previewed [here](https://htmlpreview.github.io/?https://raw.githubusercontent.com/granttremblay/astropy.github.com/master/index.html). 

Example below:


eteq:
I like it, but I wonder if you can adjust the vertical alignment so that the astropy logo is centered relative to the text instead of sticking above it?

granttremblay:
@eteq sure thing, I'll nudge it down with the CSS

kelle:
I also think it should be visually distinct as a footer. You could echo the top nav bar and make it gray shading or just add a horizontal line (like github does below 👇🏼 if you're reading this on github!)

granttremblay:
I've implemented changes requested by Erik and Kelle. Merging as we all seem to be okay with this!

PR #188, opened 2017-09-19
Update affiliated package repo with information gathered over summer

mwcraig:
This PR makes two changes:

+ Update some details of affiliated package entries that were generated while working on the numfocus affiliated package blog post this past summer.
+ Adds two new keys to the json entries: `image` and `educational_materials_link`

I can separate this into two PRs if you would like.

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -2,23 +2,25 @@
     "packages": [
         {
             "name": "specutils",
-            "maintainer": "Wolfgang Kerzendorf",
+            "maintainer": "Nicholas Earl, Adam Ginsburg, Steve Crawford",
             "provisional": false,
             "stable": false,
             "home_url": "http://specutils.readthedocs.io",
             "repo_url": "http://github.com/astropy/specutils.git",
             "pypi_name": "specutils",
-            "description": "Spectroscopic analysis utilities."
+            "description": "Affiliated package for 1D spectral operations.",
+            "image": null,
         },
         {
             "name": "astroquery",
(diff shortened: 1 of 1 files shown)


eteq:
@mwcraig - I rebased this for you because I had just  merged #185 which made a conflict.

On the extra tags, I like `image`, but am not clear on `educational_materials_link` - I guess it's not clear to me why we should have that in addition to documentation?

mwcraig:
Double check with @kelle, but I think the intent was that as packages develop more tutorial-like material the `educational_materials_link` can be a gateway to that, and can (perhaps) become an explicit target that affiliated packages should aim for.

That said, I do not think it should be *displayed* now, or perhaps only for the 1 or 2 packages that have something concrete.

eteq:
@mwcraig - understood.  My main concern is that it might be a source of confusion for new affiliated package authors. Unlike all the other terms it's not really immediately obvious what belongs there.  So I wonder if an explanation of it should go somewhere?

(Or you can split just this out in a separate PR if it requires more thought/discussion, as all the rest looks good)

astrofrog:
@mwcraig - can you rebase and split out the educational stuff for now so we can discuss it separately? The other changes look good.

mwcraig:
Will update tonight

On Wed, Sep 27, 2017 at 4:09 AM, Thomas Robitaille <notifications@github.com

mwcraig:
rebase/edits are done

astrofrog:
Thanks @mwcraig!

PR #189, opened 2017-09-20
Update Affiliated packages page: reorganize and refresh

taldcroft:


eteq:
For context to any who might come to this issue later: the removal of the "featured packages" is because those are very stale and haven't been updated in years (which was the original plan).

@taldcroft - can you clarify why you move the install instructions down, though?  The original thinking was that they should go before the table because the most important piece of info for the user is "how do I get them".  (but I'm open to persuasion, as the rest is all good)

astrofrog:
In the interest of getting this merged (I really don't want to see montage-wrapper in the featured list anymore) I am going to move the install instructions back, and we can discuss in a separate PR?

PR #191, opened 2017-09-21
Improve astropy.org landing page

taldcroft:
This simplifies the Install Astropy instructions and adds some cross-reference links.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -509,57 +509,6 @@ section.install ul li:first-child a{
 	margin-left: 0px;
 }
 
-section.install #os-selector{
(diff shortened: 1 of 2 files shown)


eteq:
@taldcroft - I added a modified version of the "additional install instructions" in b16f7a7 - I think that's important to note because it includes extra info like how to build from local source and so on.  

If you're OK with that then I think this is good to go, though!

astrofrog:
I'm taking the liberty of merging this PR since there are actually things that are incorrect in the current version (e.g. windows installers)

PR #193, opened 2017-09-28
Update mission statement

astrofrog:
This is a relatively big change - I would like to propose updating the mission statement. The original statement made it sound like we just developed a single package and just helped other packages be interoperable. This is clearly no longer the case - we as a community are coordinating the development of several packages including but not limited to the core package. Preview of the new statement:

screen shot 2017-09-28 at 13 20 20

I also think we may want to consider updating the tag line in the logo, or even removing it altogether (now that most people would recognize the actual logo bit on the left). We may actually want to have a logo that says 'The Astropy Project' instead and then have a version that has astropy in lowercase for the package (but the logo stuff could be done separately - let's focus on the mission statement here)

I wanted to get some initial feedback about these ideas, but then if there is some support for this I can post to astropy-dev since this is an important change.

### DIFF
--- a/index.html
+++ b/index.html
@@ -78,8 +78,8 @@
 
 	<section id="hero">
 		<img src="images/astropy_banner.svg" width="524" onerror="this.src='images/astropy_banner_96.png'; this.onerror=null;" alt="Astropy: a community Python library for Astronomy" />
-		<p>The Astropy Project is a community effort to develop a <a href="http://docs.astropy.org">single core
-		package</a> for Astronomy in Python and foster interoperability between <a href="affiliated/index.html">Python
+		<p>The Astropy Project is a community effort to develop a <a href="http://docs.astropy.org">common core
+		package</a> for Astronomy in Python and foster an ecosystem of <a href="affiliated/index.html">interoperable
 		astronomy packages</a>.</p>
 		<p class="acknowledge">Please remember to <a href="acknowledging.html">acknowledge</a> the use of Astropy!</p>
   </section>


pllim:
👍 for 'The Astropy Project'

granttremblay:
I like it! 

Two (very) quick ideas for a potential 'The Astropy Project' logo: 

test1-01
test2-01

pllim:
@granttremblay , I like the second one. It is more compact.

pllim:
Does this mean, we need to update the logo on https://astropy.logosoftwear.com/ products? c/c @perrygreenfield

bsipocz:
I like the 2nd one better, too, though the snake may need a bit more shrinking to better match the total height of the texts.

granttremblay:
More interations, following @bsipocz's suggestion, and trying different capitalization schemes

test3-01
test4-01
test5-01

pllim:
FWIW the third one seems more consistent.

bsipocz:
I like the 1st best, but the 3rd one is OK, too.

taldcroft:
:+1: on the tagline change.  Also :+1: on dropping the tagline in the logo.

I vote for the 3rd logo for the project, though all of them are getting typographically very busy (caps,font size, colors, boldness).  But I don't see an obvious better solution for "The Astropy Project" without punting on the current logo theme.

pllim:
@granttremblay , the very end of snake's tail seems a bit wider than the rest, which makes it looks like a rattlesnake.

perrygreenfield:
As I recall, there were oddities in the original graphic that caused issues like this when converting to other scales or representations. Caused me a lot of trouble with tshirts and stickers :-(

granttremblay:
Good catch, @pllim! 

Yup, I just confirmed that it's a bizarre oddity introduced when I imported the astropy_banner.svg (in the images/ directory) to Adobe Illustrator, which I used to make the logo prototypes above. I can manually fix it.

eteq:
👍  to the wording change.  Also 👍  to simplifying the text.  I had misgivings, but worked my way through them by admitting to myself that we've "won" and most people seeing that name now have some idea what it is :wink:.

On the logos, I think I slightly favor 3 from https://github.com/astropy/astropy.github.com/pull/193#issuecomment-332883088, but 1  is also good. 

And @granttremblay @astrofrog - just to clarify: the canonical logos go in https://github.com/astropy/astropy-logo, so we would want to put a PR there simultaneous with anything changing the logo on the web page.  Of course we can finalize the discussion in this PR, but it might just be mechanically easier for this PR to be just the wording change and have a second one (along with one in `astropy-logo`) finalizing the logo change.

eteq:
Oh, also, the README in https://github.com/astropy/astropy-logo may have the explanation to the widening-tail mystery, @granttremblay : the README there indicates that it was created in [inkscape](https://inkscape.org/), and the generated SVG might have subtle differences when imported in illustrator.

astrofrog:
I guess one thing that we'll have to make clear in astropy-logos is which logos are for the project and which for the core package (and some with just the actual galaxy are of course for either).

In any case, this specific PR doesn't need to include the logo changes. Should I send an email to astropy-dev to get feedback on the new mission statement?

astrofrog:
> subtle differences

subtle... or not

eteq:
Alright, I'll merge this as-is but create a new issue (and assign @granttremblay) about the new logo.

PR #195, opened 2017-10-02
Added astropy-healpix to affiliated package list

astrofrog:
We are applying for affiliated package status - the email to astropy-dev will be sent later today.

We think this is core enough that this could be considered for merging into the Astropy core package, or if not, become an astropy 'coordinated' package.

(this should only be merged once the package has been reviewed)


astrofrog:
The package was reviewed and accepted on November 27th 2017: https://github.com/astropy/astropy-healpix/issues/69

PR #196, opened 2017-10-02
Add reminder about updating astropy-core-maintainers group

taldcroft:
Per discussion at the coordinators meeting today, we want to add a reminder that any update to roles needs to be reflected in the `astropy-core-maintainers` google group.  This list is now meant to include all those who have a role on the team page.


eteq:
LGTM - merging!

PR #197, opened 2017-10-03
Mention slack on website

astrofrog:
This replaces gitter by slack on the website. Just to explain the motivation for the change in wording, I feel like previously we made it seem too much like people would actually get an immediate reply, but I think we want to make it very clear this is not the case and that people shouldn't get frustrated if we don't reply immediately (when people use IM services they have waaaaay less patience than email lists). I'm open to suggestions though.

### DIFF
--- a/contribute.html
+++ b/contribute.html
@@ -80,8 +80,17 @@
 	welcome and encouraged. These should ideally
 	be posted to our <a href="http://astropy.userecho.com/">UserEcho Forum</a>. </li>
 
-	<li>If you have feedback you would prefer to keep private, you can e-mail <a href="mailto:feedback@astropy.org">feedback@astropy.org</a>. This address points to a private mailing list that includes the astropy core developers. If you would like a reply (e.g., an acknowledgement of your comment), please request it.</li>
-	<li>For the extremely impatient, astropy developers often can be found in <a href="https://gitter.im/astropy/astropy">Astropy's gitter channel</a>. Gitter is basically a live web chat, but with features that make it better for discussing code. If you prefer using IRC clients for chat, there is <a href="https://irc.gitter.im/">a gitter to IRC bridge</a>.</li>
+	<li>If you have feedback you would prefer to keep private, you can e-mail <a
+	href="mailto:feedback@astropy.org">feedback@astropy.org</a>. This address
+	points to a private mailing list that includes the astropy core developers. If
+	you would like a reply (e.g., an acknowledgement of your comment), please
+	request it.</li>
+
+	<li>For the extremely impatient, astropy developers often can be found in the
+	<a href="https://astropy.slack.com">Astropy Slack team</a> (get an account <a
+	href="http://joinslack.astropy.org">here</a>). Slack is basically a live web
+	chat.</li>
+
 	</ul>
 
 	</section>
--- a/help.html
+++ b/help.html
@@ -77,7 +77,8 @@
 		please message one of the <a
 		href="https://www.facebook.com/groups/astropython/admins/">admins</a> with a
 		brief description of your involvement in astronomy. This is a very active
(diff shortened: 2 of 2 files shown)


help.html
```diff
@@ -60,19 +60,43 @@
...
-		<li><a href="https://twitter.com/astropy">@astropy on Twitter</a> - Mention @astropy in a tweet with your question. Your tweet may be re-tweeted in order to reach the relevant community. Many tweets get responses within a few hours.</li>
+		<li><a href="https://twitter.com/astropy">@astropy on Twitter</a> - Mention
+		@astropy in a tweet with your question. Your tweet may be re-tweeted in
+		order to reach the relevant community. Many tweets get responses within a
+		few hours.</li>
+
+		<li><a href="https://www.facebook.com/groups/astropython/">Python Users in
+		Astronomy Facebook group</a> - Post questions or start discussions on
+		anything related to Python programming applied to astronomy. This is a
+		closed group. If your request to join is not approved after a couple days,
```
bsipocz: can we somehow say that this is probably the most helpful channel for beginners?

contribute.html
```diff
@@ -80,8 +80,17 @@
 	welcome and encouraged. These should ideally
 	be posted to our <a href="http://astropy.userecho.com/">UserEcho Forum</a>. </li>
 
-	<li>If you have feedback you would prefer to keep private, you can e-mail <a href="mailto:feedback@astropy.org">feedback@astropy.org</a>. This address points to a private mailing list that includes the astropy core developers. If you would like a reply (e.g., an acknowledgement of your comment), please request it.</li>
-	<li>For the extremely impatient, astropy developers often can be found in <a href="https://gitter.im/astropy/astropy">Astropy's gitter channel</a>. Gitter is basically a live web chat, but with features that make it better for discussing code. If you prefer using IRC clients for chat, there is <a href="https://irc.gitter.im/">a gitter to IRC bridge</a>.</li>
+	<li>If you have feedback you would prefer to keep private, you can e-mail <a
```
eteq: You know PEP8's 80 line rules don't apply to HTML, right? :trollface:

help.html
```diff
@@ -60,19 +60,43 @@
...
+		contributors are using Slack (which has a live web chat format)
+		for everyday discussions, so please do feel free to drop by if you have
+		questions about using or contributing to astropy. Be aware that while this
+		is a live chat forum, depending on the time of day/year there may
+		not necessarily be someone available to help, so if you don't get an answer
+		here, be sure to try out one of the other methods of communication!.</li>
+
+		<li> <a href="http://groups.google.com/group/astropy-dev">Developer Email
+		List</a> - Start discussion and ask questions about changing Astropy's
+		behavior or adding new features.</li>
```
eteq: Maybe also mention this is where significant announcements for contributors/developers are usually made? Also maybe point out that this is also where *decisions* get made, not just question-asking.

contribute.html
bsipocz: Oh, but that's because html is ... (CoC) ..., right? :trollface:

astrofrog: 1x55na

astrofrog: I PEP8'd some fortran recently

MSeifert04: 80 line width isn't actually PEP8, if I [understand it correctly it's a punch card thingy](https://softwareengineering.stackexchange.com/a/148678/198787).

help.html
astrofrog: Done

help.html
astrofrog: Done

astrofrog:
@bsipocz @eteq - please re-review and let me know if we are good to merge!

astrofrog:
Since @taldcroft and @bsipocz have approved, I'll go ahead and merge!

PR #209, opened 2017-11-08
Remove mention of provisional affiliated packages

astrofrog:
I know this page will need to be updated in future but this is just a temporary change to avoid confusion.


eteq:
I think this is reasonable as a stopgap until we get the new stuff up-and-running... But what about `spherical_geometry`?  That's the one remaining provisional package on the current list.  With this change it just drops off the list, I believe?  I can add a commit to update the javascript, @astrofrog, but just want to make sure that wasn't intended...

eteq:
To follow-up here for posterity - @astrofrog and I agreed this was fine for now, because in the "new" page it will come back, just with some "red lights"

PR #211, opened 2017-11-08
Added reviews to registry.json

astrofrog:
@eteq @taldcroft @kelle - this adds the published reviews to the affiliated package registry. This then becomes the places where authoritative reviews are stored.


eteq:
@granttremblay - can you have a look at this and just make sure there's nothing about it that seems like it would cause trouble based on your initial draft new page?

astrofrog:
Just for info, I've also switched the CI from Travis to CircleCI, which is way faster and so far hasn't had any backlog issues.

affiliated/validate_registry.py
```diff
@@ -0,0 +1,89 @@
...
+
+        difference = REVIEW_KEYS - set(review.keys())
+        if difference:
+            raise ValueError(f"Missing review key(s) for {name}: {difference}")
+
+        for key, value in review.items():
+
+            if key == 'functionality':
+                if value not in REVIEW_FUNCTIONALITY:
+                    raise ValueError(f"Invalid functionality in review for {name}: {value}")
```
eteq: Here (and the two other similar ones below), maybe the error should end in ``...{value}. Valid options are: {REVIEW_FUNCTIONALITY}.``?

eteq:
One thing to *consider*, @astrofrog: add a "pre" script that checks for changes to the affiliated machinery.  Right now this runs on *every* PR regardless of whether it touches the registry.  This probably requires a CI-specific script though.  Do you know any easy way to have CircleCI report its diff? Could just check if affiliated.json is in that and pass automatically if not.  CircleCI 2.0 has some much fancier workflow machinery, but a 5-minute browse of the docs didn't suggest to me an easy way to do this other than writing our own script...

(this could be a separate issue if you don't know how to do it easily as part of this PR, @astrofrog)

astrofrog:
@eteq - I can't figure out an easy way to do this either - but CircleCI takes ~30 seconds to run and starts up almost instantaneously, so I don't think it's a problem?

eteq:
@astrofrog - fair enough.  I'm more concerned about if the tests are failing in master and other PR's see alse failures... But of course then we just ignore it.  I'll make an issue about it but agree it's a low priority.

taldcroft:
I did not check the content for all packages, but the format looks good to me.

PR #216, opened 2018-01-10
Change GSoC coordinator lead to Brigitta Sipocz

taldcroft:


eteq:
This raises a question of how we want to have the "old" coordinator names kept - clearly this was meant to have a 2017 and 2018 section, although I agree with the change here that it doesn't make sense to treat GSoC different from anything else.  But somehow/somday we should start keeping records on these roles that are reasonably accessible...

PR #217, opened 2018-01-12
Adding dust_extinction package

karllark:
Add dust_extinction package to the astropy affiliated file


eteq:
updated with review.  Can update the review if anything in karllark/dust_extinction#72 changes

PR #222, opened 2018-02-12
Add 3.0 release announcement

eteq:
This PR adds text for the release announcement for Astropy v3.0.

One thing to note: The "cite this" part is slightly awkward given paper ii is not yet in the journal.  I basically adopted @bsipocz's approach from astropy/astropy#7158, although modified to fit the html.  There is an open question whether we should update the web site's citation text (http://www.astropy.org/acknowledging.html) to *also* say this.  I'm inclined to wait on that until the paper is actually in the journal.  If we agree on this I'll just make a WIP PR to do that separate from this one.  Any opinions on this, @adrn, @crawfordsm, @hamogu, or @bsipocz?

### DIFF
--- a/acknowledging.html
+++ b/acknowledging.html
@@ -67,12 +67,16 @@
 
 		<h3>In Publications</h3>
 
-		<p>If you use Astropy for work/research presented in a publication (whether directly, or as a dependency to another package), we ask that you cite the <a href="http://dx.doi.org/10.1051/0004-6361/201322068" target="_blank">Astropy Paper</a> (<a href="http://adsabs.harvard.edu/abs/2013A%26A...558A..33A" target="_blank">ADS</a> -
-		<a href="http://adsabs.harvard.edu/cgi-bin/nph-bib_query?bibcode=2013A%26A...558A..33A&amp;data_type=BIBTEX&amp;db_key=AST&amp;nocookieset=1" target="_blank">BibTeX</a>).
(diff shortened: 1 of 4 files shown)


astrofrog:
Maybe you can also update the link to the what's new and latest version on the front page in this PR?

bsipocz:
I think we should mention the new paper in this announcement as this seems to be the best way to let our users know.
However I don't have strong opinion when to update the acknowledgement page.

hamogu:
I don't have a strong opinion either, but a slight preference to update the acknowledgement page at the same time. In my workflow, I would read the announcement email and then delete the email (yes, I am so old that I delete emails), but remember in the back of my head that something was going on with the acknowledging. When I need it next time, I might remember to check it out on the webpage - so it should be on there, too.

Plus, the ADS will track citations to astro-ph and merge them with the citations of the journal paper, once that is available, so people will start racking up credit. 
I'd be weary if we had a questionable referee report, but the current report is pretty straight forward and I think we gave enough discussion opportunity in the issues, so we (in some cases I) should move soon to get that addressed. The final paper will likely not be that much different from the version that's currently on astro-ph.

bsipocz:
@eteq - The easiest was to directly push the new contributors and version number updates to this PR rather than open a separate one. I hope you don't mind.

eteq:
Alright, I'm ok with that @hamogu  - I'll just update the citation page to exactly match this announcement in this PR

eteq:
I think this is ready to go as soon as we fix the RTD builds and have things up and running on conda

eteq:
It's now in conda and the docs pages are up. So merging!

PR #225, opened 2018-02-20
Fix #194: Simplify role-changing machinery with json and js

### LINKED ISSUE #194: Simplify role-changing machinery with json and js
Right now it's a bit awkwardly complicated to update the roles - you have to update a special-format ``roles.txt`` file, and then run a script locally and remember to commit both with git.  That works for now, but A better system might be to:

- [ ] change ``roles.txt`` to a ``roles.json`` file
- [ ] use code like what we have in the affiliated package page to populate a table *from* the JSON using a javascript function (which would run locally and is very fast - this has been in practice very stable/easy on the ``affilaited/index.html``).
- [ ] include the roles *descriptions* in the``roles.json`` as well - then do a bit more js tricks to get the roles list to appear on the web site.

Once these are all done, it'll be easy to update the roles in a single commit rather than the awkward 2- or 3- stage process currently required (which also requires knowing to run the script).

Also #167 could be much more easily addressed with this all in place - there could be an additional json field for past holders of the role, and the js could include that at the bottom of the page or something.

apooravc:
Fixes #194 

- Deleted `roles.txt` and `update_roles.py`.

- Added `roles.json`, an equivalent of `roles.txt`.

- Used some JS to populate roles table from the json file.

- Added roles descriptions to json file and created those roles lists from it. 

_json entry format_
{
      "role" : "role name",
      "url" : "div element url",
      "sub-role" : ["sub-role-1", "sub-role-2", "sub-role-3"],
      "lead" : ["lead-1", "lead-2", "lead-3"],
      "deputy" : ["deputy-1", "deputy-2", "deputy-3"],
      "role-head" : ["role heading"],
      "strong" : [["strong-title-1"], ["strong-title-2"]],
      "sub-description" : [["sub-description-1"], ["sub-description-2"]],
      "description" : ["list-item-1", "list-item-2", "list-item-3"]
}

One way to see the changes _locally_ would be as follows:
1) Clone my repo from [https://github.com/apooravc/astropy.github.com](https://github.com/apooravc/astropy.github.com)
2) After cloning, only the master branch would be available locally. Enter `git checkout -b simplify-role-changing origin/simplify-role-changing ` to get the branch _simplify-role-changing_ locally.
3) The changes could then be seen on team page.

@eteq Please have a look.


js/functions.js
```diff
@@ -20,9 +20,127 @@ $( document ).ready(function(){
...
+            if (role["description"].length != 0) {
+                var list = '';
+                //generate list by iterating over each description in the role
+                if (index != 5) {
+                    role["description"].forEach(function (description) {
+                        list += '<li>' + description + '</li>';
+                    });
+                    $(".roles-list").eq(index).append(list);
+                } else {
+                    //separate case for index 5 representing tutorials & guides section
```
eteq: Maybe modify this to not be specific to a particular index?  E.g., instead of the if statement being `if (index != 5)` have it be `if (role["description"].length == 2 & role["description"][0] instanceof Array)`?  (if you use that `if` you'll have to swap the if and else blocks, but hopefully you get the idea?)

eteq:
(One other minor suggestion for the future @apooravc : usually it's best to have commit messages be more specific than "fix".  For example, instead of "Third fix for #194" you might say "added urls to javascript-generated table".  No change needed, but just something to keep in mind for the future.)

apooravc:
@eteq Did the required changes, please have a look.

js/functions.js
```diff
@@ -20,9 +20,135 @@ $( document ).ready(function(){
...
+
+    function createRolesTable(roles) {
+        //roles is an array of objects called "role"
+        var rows = '';
+        roles.forEach(function (role) {
+            //role is an object containing information about each team role
+            //index marks current lead
+            var index = 0;
+            //regular expression used in searching for 1 in a string
+            var regExp = /^1/i;
```
eteq: this `regExp` variable could have a slightly clearer name - I was trying to understand and it took me a while to realize it was the "deputy" part.  Maybe call it `regExpPreferDeputy` or similar?

roles.json
```diff
@@ -0,0 +1,153 @@
+[
+    {
+      "role" : "Coordination committee member",
+      "url" : "Coordination_committee_member",
+      "sub-role" : [],
+      "lead" : ["Tom Aldcroft", "Kelle Cruz", "Tom Robitaille", "Erik Tollerud"],
+      "deputy" : [],
+      "role-head" : ["Coordination committee member (4 people)"],
+      "strong" : [],
+      "sub-description" : ["Overall coordination and management of the Astropy project, including:"],
```
eteq: I think the `description` and `sub-description` or swapped here.  What you have as `sub-description` seem to be the *more minor* items, whereas `description` are the big things.  Maybe just swap the two?

eteq:
With a day to think I realized I could just do those minor changes myself.  So this is all set and I will merge it shortly.  Thanks @apooravc !

PR #227, opened 2018-02-22
Clarify code of conduct inheritance

astrofrog:
As discussed with @kelle, @taldcroft, and @eteq

### DIFF
--- a/code_of_conduct.html
+++ b/code_of_conduct.html
@@ -39,14 +39,14 @@
                     <ul>
                     <li><a href="about.html">About Astropy</a></li>
                     <li><a href="code_of_conduct.html">Code of Conduct</a></li>
-                    <li><a href="acknowledging.html">Acknowledging</a></li>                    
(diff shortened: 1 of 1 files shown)


code_of_conduct.html
```diff
@@ -65,28 +65,74 @@
...
+		treated with the strictest confidence) or talking privately with the
+		person.</li>
+
 		</ul>
 
-		<p>This code of conduct applies to all community situations online and offline, including mailing lists, forums, social media, conferences, meetings, associated social events, and one-to-one interactions.</p>
+		<p>This code of conduct applies to all community situations online and
+		offline, including mailing lists, forums, social media, conferences,
+		meetings, associated social events, and one-to-one interactions.
+		Projects within Astropy, such as for example affiliated packages or
```
eteq: Maybe have this be a separate paragraph? (i.e. `<p></p>`)

eteq: Also, maybe drop "such as" in this line, since it already says "for example"?  (or drop "for example")

PR #212, opened 2017-11-08
Add synphot to affiliated package registry

pllim:
Add `synphot` to affiliated package registry.

PR #232, opened 2018-03-02
Registry clean up

pllim:
I moved the OCD portion of #212 to here. Some clean-up to ensure all entries are consistent.

p.s. A glance at `linetools` doc indicated to me that it should not be listed as "stable" but that is not my call to make. FYI.


pllim:
(I think the CircleCI failure is due to their "partial system outage")

affiliated/registry.json
```diff
@@ -595,9 +596,11 @@
             "maintainer": "Adam Ginsburg <adam.g.ginsburg@gmail.com>",
             "provisional": false,
             "stable": true,
-            "home_url": "https://pyspeckit.readthedocs.org",
+            "home_url": "https://pyspeckit.readthedocs.io",
             "repo_url": "https://github.com/pyspeckit/pyspeckit",
             "pypi_name": "pyspeckit",
+            "description": "An extensible spectroscopic analysis toolkit for astronomy.",
+            "image": null
```
eteq: missing comma here

affiliated/registry.json
```diff
@@ -316,11 +316,11 @@
...
             "provisional": false,
             "stable": false,
-            "home_url": "http://gwcs.readthedocs.io",
-            "repo_url": "https://github.com/spacetelescope/gwcs.git",
+            "home_url": "https://gwcs.readthedocs.io",
+            "repo_url": "https://github.com/spacetelescope/gwcs",
             "pypi_name": "gwcs",
             "image": null,
-            "description": "Package for managing the World Coordinate System (WCS) of astronomical data.",
+            "description": "Package for managing the generalized World Coordinate System (WCS) of astronomical data.",
```
eteq: I think this should *not* be changed, because it implies that "World Coordinate System" as a term is *not* general, which it is.  So I suggest un-capitalizing it (as that implies "FITS-WCS") but leaving off the generalized

pllim:
There *is* a "G" in GWCS... But I am okay with your changes. :+1:

eteq:
@pllim - yeah, but I think that doesn't apply to the *concept*, rather the package.  Anyway, I see your point, so I'm going to make a PR to update the *docs* of GWCS to address this head-on.

PR #236, opened 2018-03-21
Add roles json validator

eteq:
This PR adds another validator to use with circleCI, inspired by the affiliated package checker (thanks @astrofrog!). This one checks that the roles json file is valid. It also includes some minor improvements to the affiliated package checker based on ideas I had while writing this one.

One thing I wasn't sure about - should this be a separate `run` step in circleCI or a separate job?  I chose the former because it seemed easier, but perhaps @astrofrog can advise?

(**NOTE**: the last three commits *intentionally* break things in several ways and then reverts them to see how the CI responds.  Before merging these should be removed with a rebase!)


eteq:
Rebased.  Will wait for the CI  just to make sure I didn't muck anything up in the circle CI config.

PR #239, opened 2018-04-23
Update versions

bsipocz:
There are issues with RTD so the docs for the new releases are not yet build. Do not merge this until RTD is all clear.


bsipocz:
The RTD builds are ready, so I'm merging this now. The circleCI failure seems unrelated but real.

PR #241, opened 2018-05-07
A new design for the Affiliated Packages Table

granttremblay:
Partially in the spirit of astropy/astropy-APES#40, @eteq (LONG ago) proposed an overhaul / refresh of the Affiliated Packages table. 

Following on #pyastro18's recent Hack Day, @eteq, @astrofrog and myself propose the following design. Note that this specific design is entirely the brainchild of @astrofrog, and @eteq did most of the actual implementation. 

screen shot 2018-05-07 at 11 02 07 am

This new table aims to simultaneously increase per-package information density while also improving readability. Miraculously, I think it does both (thanks entirely to @astrofrog and @eteq's great ideas). 

This table, of course, dynamically pulls from the JSON registry, and can be easily adapted to any forthcoming evolution of the Astropy affiliated package model. 

Edit: You can see a live preview of this change [here](http://hea-www.cfa.harvard.edu/~tremblay/Astropy_Preview/preview/affiliated/index.html).

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -357,10 +357,10 @@ var review_name_map = {"functionality": "Functionality",
     "python3": "Python 3"
 };
 
-var review_default_color = "green";
+var review_default_color = "brightgreen";
 var review_color_map = {'Unmaintained': "red",
     "Functional but low activity": "orange",
-    "Good": "green",
+    "Good": "brightgreen",
     "Partial": "orange",
     "Needs work": "red"
 };
--- a/js/functions.js
+++ b/js/functions.js
@@ -225,16 +225,26 @@ function url_translator(urltext) {
     if (urltext === undefined) {
         return 'None';
     } else {
-        return '<a href="' + urltext + '">' + urltext + '</a>';
+        return '<a href="' + urltext + '">' + 'Website' + '</a>';
     }
 }
+
+
+function repo_translator(urltext) {
+    if (urltext === undefined) {
+        return 'None';
+    } else {
+        return '<a href="' + urltext + '">' + 'Repository' + '</a>';
+    }
+}
+
 
 function pypi_translator(pypiname) {
     if (pypiname === undefined) {
         return 'None';
     } else {
         var urltext = 'http://pypi.python.org/pypi/' + pypiname;
-        return '<a href="' + urltext + '">' + pypiname + '</a>';
+        return '<a href="' + urltext + '">' + 'PyPI' + '</a>';
     }
 }
 
@@ -250,6 +260,7 @@ var _email_regex_str = '[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}';
 var _email_regex  = new RegExp(_email_regex_str, 'i');
 var _email_with_name_regex  = new RegExp('(.+)<(' + _email_regex_str + ')>', 'i');
 
+
 function maintainer_translator(maint, pkgnm) {
     var url, match;
     if (_email_with_name_regex.test(maint)) {
@@ -264,24 +275,13 @@ function maintainer_translator(maint, pkgnm) {
     }
 }
 
+
 function populateTables(data, tstat, xhr) {
-    populateTable('accepted-package-table', data, false);
-    populateTable('provisional-package-table', data, 'only')
+    populateTable('accepted-package-table', data);
 }
 
-function populateTable(tableid, data, allowprovisional) {
-    // First we use allowprovisional io decide what the checkProvisional function (used below) does
-    var checkProvisional;
-    if (allowprovisional === false) {
-        checkProvisional = function(provisional_status) { return provisional_status === false; }
-    } else if (allowprovisional == 'only') {
-        checkProvisional = function(provisional_status) { return provisional_status; }
-    } else if (allowprovisional === true) {
-        checkProvisional = function(provisional_status) { return true; }
-    } else {
-        throw "Invalid allowprovisional value " + allowprovisional;
-    }
 
+function populateTable(tableid, data) {
     // Now we get the table and prepare it
     var tab = document.getElementById(tableid);
     var ncols = tab.rows[0].cells.length;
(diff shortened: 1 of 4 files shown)


astrofrog:
@granttremblay - nice! Can we change the shields from green to brightgreen?

astrofrog:
Another quick request - what about including some kind of faint grey line between entries to have a slightly clearer separation?

granttremblay:
@astrofrog Done! 
screen shot 2018-05-07 at 5 03 50 pm

granttremblay:
Whoops! Line placement was messed up in that last commit - sorry it squeaked past. 

Fixed now: 
screen shot 2018-05-07 at 5 08 05 pm

taldcroft:
Can we see the whole thing live somewhere?

taldcroft:
I just tried to load this locally (`open index.html`) from the granttremblay/master branch and it just hangs with `Loading ...` in the affiliated packages section.

granttremblay:
@taldcroft I've placed a temporary live preview of my branch here: http://hea-www.cfa.harvard.edu/~tremblay/Astropy_Preview/preview/affiliated/index.html

It indeed won't work locally if you simply load it up in (e.g.) Chrome, I *think* because of [this issue](https://jamison.dance/post/12857366951/running-local-javascript-files-in-chrome).

astrofrog:
@granttremblay - looks nice! Could you make the font smaller to match the font outside the table? On my computer the font in the table is much larger than the text before the table.

granttremblay:
Sure thing. I had originally made the text 40% larger than outside the table, to match the mockup, but I can definitely make it smaller. A little bit bigger might look a *bit* better, but 0% size bloom will be fine if I shrink the padding a bit. 

Matching text size: 

screen shot 2018-05-08 at 10 57 57 am


20% boost in size: 

screen shot 2018-05-08 at 10 58 07 am

granttremblay:
[Affiliated Package Table Overhaul](https://trello.com/c/phjohFGZ/2-affiliated-package-table-overhaul)

granttremblay:
[Affiliated Package Table Overhaul](https://trello.com/c/phjohFGZ/2-affiliated-package-table-overhaul)

eteq:
Looking good, @granttremblay !

Re: @taldcroft's
> it just hangs with Loading ... in the affiliated packages section.

I had the same problem @granttremblay says with Chrome - the most expedient workaround I've found is to just use Safari for local testing (on a mac) - by default it has the same policy of not loading local files as in the link @granttremblay mentioned, but you can turn it off in the developer tools (whereas chrome requires a command-line argument/restart).  I think there's a similar option in firefox but I can't remember exactly where.


Re: @astrofrog's size suggestion, Seeing these side-by-side I think matching size is best.  And actually, I wonder if it would look better if the table were actually a bit *smaller*? (say 90% or 80%?)   That makes it clearer that the table is sort of "detailed" information, and the main text is the page content.

js/functions.js
```diff
@@ -264,34 +275,22 @@ function maintainer_translator(maint, pkgnm) {
...
-        checkProvisional = function(provisional_status) { return provisional_status === false; }
-    } else if (allowprovisional == 'only') {
-        checkProvisional = function(provisional_status) { return provisional_status; }
-    } else if (allowprovisional === true) {
-        checkProvisional = function(provisional_status) { return true; }
-    } else {
-        throw "Invalid allowprovisional value " + allowprovisional;
-    }
 
+function populateTable(tableid, data) {
```
eteq: Quick thing to note (no change required): in a follow-on PR we'll probably want to add back in something like the `allowprovisional` argument - we'll want to separate out the tables for the "affiliated" and "coordinated" astropy packages.  But that should be in a later PR because we still need to finalize how to harmonize this with the "team" listings (some of which requires coordination committee role-assignments).

saimn:
Another way (easier, imo) to test this locally, is to spawn a web server, e.g. with `python -m http.server`.

taldcroft:
About font size, I agree with @eteq that matching size is a good starting point, but it might be worth trying a smaller font (but I suspect that matching size is probably cleaner from the UI perspective).

One big picture design comment is that I think having only one giant "detail" table in the middle of the doc is problematic in two ways:
- Page usability: useful information is lost at the end of the page.
- Package discovery by users: it's just hard to use this table to find the package you need because you have to scroll through and evaluate a lot of content that is mostly irrelevant (for discovery).

Imagine in a few years if there are a hundred or more affiliated packages.  What I think is needed is a summary table that has two columns, package name and a short one-line summary description (which would be written with discovery in mind, i.e. loaded key words and terse). The package name would link to the detail row in the full table.

Thoughts @adrn (since you had opinions at the dev meeting)?

taldcroft:
One technical issue is that the new table renders poorly on mobile in the sense that it doesn't fit and one has to scroll horizontally to see the description.

bsipocz:
@taldcroft - I prefer the other way around, having the detailed table as a default, but making it easy to get to the basic table either by clicking somewhere or with using some JS to hide the details.

bsipocz:
With the current design I have issues with the website. Most packages doesn't have a standalone website like astropy.org/sunpy.org, only their documentation. So I think it would be more useful to link to `Documentation`, `Repository`, and `PyPI`

granttremblay:
Following a quick chat with @taldcroft, I've reduced the table font size to match that of the main site, and slightly reduced the table size padding. I had been holding off on doing this given the (excellent) issues raised by @bsipocz and @taldcroft above, but it seems as though the coordination committee is happy with addressing those via separate PRs. 

[**Live Preview of the new Affiliated Packages Table**](http://hea-www.cfa.harvard.edu/~tremblay/Astropy_Preview/preview/affiliated/index.html)

Remaining important issues (to be addressed by separate PRs): 

- Improve responsive reflow of table for mobile users. The table looks terrible on a mobile screen. It's tolerable on a Tablet screen. 
- Think about a more general set of links - rather than linking to a package-specific website - as many such packages don't have standalone websites beyond their RTD page. 
- The table makes for a pretty lengthy scroll. As Astropy expands, this will only get worse. We could think about perhaps collapsing the table into a minimal matrix of buttons by default, and using an "expand" button to blow up the table (or individual packages) to show the information we now currently show by default. 

In any case, this might be ready for a merge. Will wait for approval by Coordination Committee members before doing so.

PR #253, opened 2018-07-24
Replace Tom with Thomas

pllim:
To be consistent with name listed under "Core Package Contributors". cc @astrofrog


bsipocz:
Given the approval, I'm merging.

pllim:
Tom A, since your name is consistent in both sections, you are off the hook! :wink:

pllim:
For those who are curious, I was data scrapping off that page and the inconsistency gets in the way. :sweat_smile:

PR #259, opened 2018-08-22
Add more informative text for MARXS

hamogu:
After all, the whole point of displaying a description is that that description should tell the reader on first glance if it's worthwhile to read the full documentation.


bsipocz:
Thanks @hamogu!

PR #265, opened 2018-09-22
Simplify citation information on acknowledgements page

adrn:
1. This changes "Acknowledging" to "Acknowledging & Citing" in the menu bar
2. I tried to simplify the suggested citation and text surrounding the topic, and added a button to copy the bibtex entries for both papers to the clipboard (similar to ADS)

This fixes #263 

cc @astrofrog @eteq

PR #266, opened 2018-10-01
Should the acknowledgement for Astropy say *both* papers or not?

eteq:
This is a carry-over from #265.  Right now the wording on the acknowledging&citing page is "If you use Astropy for work/research presented in a publication (whether directly, or as a dependency to another package), we ask that you please cite the Astropy papers" and shows both papers.

But is that really what we want? This might be a good choice to improve "credit" for Astropy, but for "reproducibility" we really want people to cite whichever one is relevant for their work - e.g. the one closer in time to the version they are using.  This PR updates the wording to reflect that... but it shouldn't be merged until we have consensus on this.

cc @adrn @astrofrog @kelle @bsipocz

### DIFF
--- a/acknowledging.html
+++ b/acknowledging.html
@@ -112,7 +112,7 @@ Bdsk-Url-1 = {https://doi.org/10.3847/1538-3881/aabc4f}}`;
 
 		<h2>In Publications</h2>
 
-		<p>If you use Astropy for work/research presented in a publication (whether directly, or as a dependency to another package), we ask that you please cite the Astropy papers:
+		<p>If you use Astropy for work/research presented in a publication (whether directly, or as a dependency to another package), we ask that you please cite one of the two Astropy papers as relevant for which Astropy version you used:
             <ul>
                 <li><a href="https://arxiv.org/abs/1801.02634" target="_blank">Astropy Paper II</a>
                     (<a href="http://adsabs.harvard.edu/abs/2018arXiv180102634T" target="_blank">ADS</a> -


bsipocz:
I'm OK with asking to cite only one of the papers, but if it's only one maybe we should point out that paper No1 is for v0.2, so keeping status quo and citing that one only is not what we prefer unless of course version v0.2 was used all the way for that given paper...

adrn:
I'm 👎on this: 
1. I think we should maximize credit
2. People don't want to read instructions to cite - the feedback I got (and the reason for the last PR) was that once people find this page, they want to copy-paste something, not have to read instructions and decide. 
3. I don't see an argument for reproducibility. Even the recent astropy paper is out of date from that perspective! The paper versions are not tied to software versions. 

That said, I would be open to removing the first paper from the recommended citation and instead adding a citation to the latest zenodo record (that would probably be most sensible for reproducibility), but then we'd have to ask that they check what version they're using and update the software citation if it's not the most recent...

bsipocz:
@adrn - my point is that it's not good to only cite the first paper for reasons we wrote a second one.

 Not sure what you mean by maximize credit, but I'm OK with providing an easy copy-pasting option.

astrofrog:
For reproducibility, what we do need is to archive specific versions of Astropy on Zenodo in case people want to cite specific versions (while still citing the paper).

But I think we should just ask people to cite both papers in terms of overall credit.

astrofrog:
Just to be clear, we don't have Zenodo entries yet but I'll work on adding those soon (we need to be careful how this is done as we want to use the new versioning capabilities on Zenodo)

adrn:
Sounds good! Should I wait until we have the zenodo entries setup then?

astrofrog:
@adrn - I think so. I'll try and do the zenodo entries within the next couple of days.

astrofrog:
@adrn @eteq - I've uploaded all existing astropy releases to Zenodo. The overall 'concept' DOI is:

https://doi.org/10.5281/zenodo.1461536

This points to whatever the latest uploaded version is and is what's used to cite the package on Zenodo but not for a specific version.

On the right there is a list of versions and each one is a Zenodo entry with its own DOI:

screen shot 2018-10-13 at 15 13 01

So to cite 3.0.4 specifically, one would use:

https://doi.org/10.5281/zenodo.1461593

adrn:
Thanks @astrofrog! Any idea why the 3.1 versions haven't uploaded?

adrn:
Oh, oops, I thought it was automatic! I didn't mean to sound passive aggressive

adrn:
I think after discussion, we decided that it is not a big ask to request that uses cite both papers, and so I'm going to close this PR. If you disagree, @eteq, please reopen, but I think @astrofrog, @kelle and I all agree that it doesn't hurt! Note that the discussion veered towards discussion zenodo DOI's - let's move any discussion about that to #291

PR #271, opened 2018-10-19
Update versions

bsipocz:
We may want to wait with this until the conda default packaging is ready.


bsipocz:
Releases are on default conda, few more cleanups to the docs and we're all set.

PR #275, opened 2018-12-03
Turn Matt Craig blue for nddata

pllim:
As requested during Astropy Coordination Meeting 2018. Is this what you wanted, @mwcraig ?


mwcraig:
It sounds like we are moving away from blue people on the list, so for we could close this.

PR #276, opened 2018-12-03
Add David Shupe as units deputy

pllim:
As proposed at Astropy Coordination Meeting 2018 by @adrn . cc @stargaser


adrn:
This was my suggestions but didn't discuss with @stargaser so I don't think this should be an official request until David requests to step up!

pllim:
Yeah, sorry. It was bad call on my end. I apologize for any inconvenience and confusion caused, and I take full responsibility.

PR #279, opened 2018-12-04
Add uncertainties roles

eteq:
Once v3.1 is released we should merge this since it adds the uncertainties subpackage.

@astrofrog and @mhvk - I added you as deputies given your heavy involvement in the dev. But if you don't want me to do that (or really want to be lead) feel free to say so!


mhvk:
I do want my `Variable` class there at some point, so happy to be sub-deputy.
(Aside: that source file is damn hard to parse!)

adrn:
As someone who just made other edits to the roles file, I agree with @mhvk -- very hard to parse and edit!

eteq:
@adrn @mhvk - if you think this is bad you should have tried to update it in the pre-js version! :wink: But your point is well taken. It might be just a "use whitespace-and-linebreaks-effectively" solution?

PR #280, opened 2018-12-04
Improve the description of pyspeckit

keflavich:
Small change to the affiliated package registry.


eteq:
Note this *also* updates the `specutils` review status, but that's intended (just not in the description).

PR #284, opened 2018-12-05
Update roles based on breakout session discussions at astropy 2018

adrn:


adrn:
I got OK's from @mwcraig and @crawfordsm in person that they are OK taking on the Twitter roles.

roles.json
```diff
@@ -44,7 +44,7 @@
       "url" : "Learn_coordinator",
       "sub-role" : [],
       "lead" : ["Kelle Cruz"],
-      "deputy" : ["Unfilled"],
+      "deputy" : ["Adrian Price-Whelan"],
```
eteq: After looking at this again it occurs to me that one could argue that this isn't really supposed to have a deputy, in that the "sub-roles" are really the deputies.  That is, there aren't "deputy" astropy coordinators because the other roles are sort of deputies.

I don't have a strong opinion, and am happy to defer to @kelle if she wants a specific deputy (and @adrn is certainly great for the role). But perhaps we should have the discussion in a separate PR?

```suggestion
      "deputy" : [],
```

adrn: Sounds reasonable to me!

PR #297, opened 2019-01-17
Add Zenodo DOI to recommended citation

### LINKED ISSUE #291: Update citation request to include zenodo DOI
Daina said citing the paper is insufficient. We also need people to cite the actual code using the Zenodo DOI.

adrn:
I need to adjust the text a little bit, so this is still a WIP, but I wanted to open this to start some discussion.

Currently in this PR, I added the bibtex record for the latest Zenodo record and added the record to the "copy" button, and added a link to the latest Zenodo DOI as a badge. The badge always links to the current version, but the bibtex record would require us to manually update it whenever a new version is released. I'm thinking we might want to link to Zenodo but not automatically add the record to the "copy" button, but wanted to hear your thoughts.

Fixes #291

### DIFF
--- a/acknowledging.html
+++ b/acknowledging.html
@@ -123,7 +123,7 @@ Bdsk-Url-1 = {https://doi.org/10.3847/1538-3881/aabc4f}}`;
 
 		<h2>In Publications</h2>
 
-		<p>If you use Astropy for work/research presented in a publication (whether directly, or as a dependency to another package), we ask that you please cite the Astropy papers:
+		<p>If you use Astropy for work/research presented in a publication (whether directly, or as a dependency to another package), we ask that you please cite the Astropy papers, and recommend that you cite the Zenodo record for the version of Astropy that you use.
             <ul>
                 <li><a href="https://arxiv.org/abs/1801.02634" target="_blank">Astropy Paper II</a>
                     (<a href="http://adsabs.harvard.edu/abs/2018arXiv180102634T" target="_blank">ADS</a> -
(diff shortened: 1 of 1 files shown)


acknowledging.html
```diff
@@ -55,9 +55,19 @@
 Year = 2018,
 Bdsk-Url-1 = {https://doi.org/10.3847/1538-3881/aabc4f}}`;
 
+        var zenodo = `@misc{astropy:zenodo,
```
bsipocz: Any way to automate this? I run into the issue so many places of the zenodo entry being so far behind the actual release that it would be superb not to end up there with astropy.

astrofrog: Not easily but we should just make it part of the release process

astrofrog: I can write some instructions for the dev docs

astrofrog: Ah wait did you mean uploading to Zenodo, or updating this page?

bsipocz: I meant this bibtex here. Creating the zenodo entry is doable automatically afaik (or at least was with GH services)

eteq: Yeah I'm with Brigitta here.  The copy-and-paste mechanism seems problematic if it gets you whatever the last one is the web site actually got updated to show.

eteq:
@adrn - I like the idea of this, but @bsipocz raised a very good point that it's going to be a pain to keep the entry on this page in sync with the core release.  Three possibilities come to mind:

1. Instead of including the actual example bibtex entry on the page, include just the part that says "cite the right zenodo entry" and link to the "rollup DOI": https://doi.org/10.5281/zenodo.1461536 (and state clearly "but be sure to pick the right version").  Downside is that a lot of folks will just ignore that.
2. Include a bibtex entry but have it be intentionally broken.  I.e., insert text like `\REPLACE THIS ENTRY WITH THE RIGHT VERSION ENTRY` to *force* people to then do something about it instead of blindly pasting
3. Build some sort of javascript machinery that auto-populates the text by reading the latest Zenodo entry from https://doi.org/10.5281/zenodo.1461536 and fills in the bibtex on that basis.  Downside is that a lot of users will probably use that even though it's the wrong version.

If it were me I'd go with 1 because that seems safest and most maintainable.  But I'd be up for any of the above, what do you think @adrn ?

astrofrog:
Once people copy the acknowledgment they will use the same one until they retire so I’m in favour of 1.

pllim:
@adrn , are you still interested in wrapping this up? Thanks!

adrn:
The last Zenodo record was for v3.0.5, so astropy/astropy#10611 needs to be resolved before I can finalize this!

adrn:
Oops - accidental close. I'll leave this open until we make a decision in the other issue ^.

hamogu:
This PR is about 6 years old, predates the last astropy papers, and has conflicts. The discussion above asked to wait for https://github.com/astropy/astropy/issues/10611 which was closed about 4 years ago, so it seems that this PR is somehow stuck.

I'm closing it - you can always re-open if you are really working on it or just open a new PR.

PR #298, opened 2019-02-10
Make higher res astropy-powered PNG

adrn:
The current image linked from acknowledging is way too low-res for modern screens!


adrn:
I think this is fairly uncontroversial, so I'm going to just merge this...

PR #305, opened 2019-03-04
Remove subsection "Other astronomy packages"

### LINKED ISSUE #304: Remove "Other astronomy packages" section?
The affiliated packages page has a section "Other astronomy packages" that seems quite outdated nowadays. The kapteyn package is f.e. only for Python 2.6. These packages can't serve as recommendation anymore, at best they mislead visitors of the page.
IMO this section should be just removed.

olebole:
This section is quite outdated: The described software is no longer developed or maintained, and does not work with Python 3. Referring to that software may be misleading for new users. 
Therefore, it should be removed.

Fixes #304.

### DIFF
--- a/affiliated/index.html
+++ b/affiliated/index.html
@@ -157,15 +157,6 @@
 	</table>
 -->
 
-	<h3><u>Other astronomy packages</u></h3>
-
-	<p>This section lists some packages that are not affiliated pacakges, but are useful reference (i.e. precursors, overlapping in functionality, etc). </p>
-
-	<ul>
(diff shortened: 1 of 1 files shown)


bsipocz:
Thanks! I'll keep this open a little bit longer to give a chance for others to oppose. If no one else does it in a few days, I'll merge. 

Thank you @olebole for catching issues like this and as a result making astropy more consistent and better.

astrofrog:
I'll cc @eteq as the author of astropysics

PR #317, opened 2019-04-02
Adjust logo

eteq:
Sadly the new most excellent logo @pllim contributed in #316 cannot last forever.  I suppose it deserves a full 24 hours given that it's a bout a "special day", though, so we should wait for at least 6 hours before merging this.  And I'm definitely not saying that because a certain pair of :cat:s gave me a judgemental look demanding their fair share of fame.


pllim:
You might also want to squash it out in `master` itself (or just drop my special commits altogether) to get rid of them from the repo footprint.

pllim:
Or if you fancy some JavaScript, maybe can even have a conditional to only show special logo on special day without reverting anything! :wink:

eteq:
Ooh, tempting, perhaps the JS trick could be a project for 2020, @pllim ? :wink:

I think the squash isn't necessary though as it's ~1 per year and they aren't that big of files.  I kinda like having this in the git history anyway.

pllim:
I am just gonna drop these links here...

https://stackoverflow.com/questions/492994/compare-two-dates-with-javascript

https://stackoverflow.com/questions/17249340/conditionally-load-images-for-responsive-designs

astrofrog:
@pllim - I think the default logo should load the normal way and JS should be used to conditionally replace the image - we don't want the logo to have to be shown using JS under normal conditions.

pllim:
See #318 for **alternate solution**.

bsipocz:
I'm afraid it's time to say goodbye for the cats 😿

PR #324, opened 2019-05-19
Proposed affiliated package: MiraPy

swapsha96:
MiraPy is a Python package for Deep Learning in Astronomy. It is built using Keras for developing ML models to run on CPU and GPU seamlessly. The aim is to make applying machine learning techniques on astronomical data easy for astronomers, researchers and students.

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -1,6 +1,23 @@
 {
     "packages": [
         {
+            "name": "mirapy",
+            "maintainer": "Swapnil Sharma and Akhil Singhal <swap.sha96@gmail.com>",
+            "stable": true,
+            "home_url": "https://mirapy.org/",
+            "repo_url": "https://github.com/mirapy-org/mirapy",
+            "pypi_name": "mirapy",
+            "description": "MiraPy: A Python package for Deep Learning in Astronomy",
+            "review": {
+             "functionality": "To be filled out by the reviewer",
+             "ecointegration": "To be filled out by the reviewer",
+             "documentation": "To be filled out by the reviewer",
+             "testing": "To be filled out by the reviewer",
+             "devstatus": "To be filled out by the reviewer",
+             "python3": "To be filled out by the reviewer",
+             "last-updated": "To be filled out by the reviewer"
+        },
+        {
             "name": "specutils",
             "maintainer": "Nicholas Earl, Adam Ginsburg, Steve Crawford, and Erik Tollerud",
             "provisional": false,


astrofrog:
Thank you for proposing this package to the Astropy affiliated package ecosystem! We will be asking a member of the Astropy project to review this submission and will get back to you once we hear back from them.

taldcroft:
This package has now been reviewed for inclusion in the Astropy affiliated package ecosystem by a member of the Astropy community as well as myself, and I have synthesized the results of the review here. I apologize for the delay in getting this review carried out.

You can find out more about our review criteria in [Reviewing affiliated packages](https://github.com/astropy/astropy-procedures/blob/master/documents/affiliated_package_review_guidelines.md). For each of the review categories below we have listed the score and have included some comments when the score is not green.

### Functionality

This area is currently "red". This is based on the following issues:

- There is no way to read FITS images or other astronomy-specific formats, so general applicability
  within astronomy is not obvious.  The image loader users `cv2.imread` so it is not
  clear how easy it is to use FITS.
- The classifiers are tied to specific surveys (e.g. ATLAS) with the exception
  of the `XRayBinary` classifier, which may be more general.
- The `autoencoder` module has one encoder, `DeNoisingAutoencoder`, 
  that is potentially of broad interest, but it was hard to  determine because the 
  only example uses a function from MiraPy to load a set of images from MiraPy.

With a bit of work the package authors should be able to get this to "specialized green."

<table>
<tr>
<td width=150>Out of scope</td>
<td> Specific to one project or collaboration.</td>
</tr>
</table>

### Integration with Astropy ecosystem 

This area is "red".  Apart from one test, the package itself uses no astropy core functionality.  The package defines its own `Model1D` class and a single subclass, `Gaussian1D`, which are not at all related to the models in astropy core. The same is true of the `ParameterEstimation` class; it uses `scipy.optimize` but does not use any of the astropy modeling machinery.

I'm not sure to what extent those decisions are driven by the desire to use the autograd numpy wrapper that provides automatic derivatives of numpy arrays.

The use of keras instead astropy functionality in `autoencoder.models` seems appropriate, given the nature of the package.

<table>
<tr>
<td width=150>Red</td>
<td>Does not use Astropy or other affiliated packages anywhere where it should be possible, and/or uses other libraries instead.</td>
</tr>
</table>

### Documentation

Although there is some documentation this is rated "red". The narrative documentation covers installation, has a list of the tutorial notebooks available (with a link to a repository where they live), and the API documentation generated from automodapi.

The installation notes do not clearly call out the dependencies and it took some trial/error to get everything installed.

In general the tutorial notebooks are  inadequate to convey to a new user how to actually take advantage of the functionality in this package.  There is little to no discussion of what the various function parameters are doing, how the user might change them in different situations or adapt the notebooks to different data sets.

Even getting the notebooks to run exceeded the patience level of the reviewer: they appear to all reference datasets that exist only in particular directories on the package author's machine (e.g. `path = 'D:\MTP\ogle'` or `data_dir = '../../image-denoising-network/messier_catalog_galaxies'`), but do not provide any obvious hint about how to get those datasets.  The docs provide links to various catalog home pages, but what is needed is clear instructions on how to get the requisite data files onto local disk (or better yet code to do that).

In addition there needs to be more explanation of the astronomy content being illustrated in each notebook and a summary of what the results at the end mean.

For example, the OGLE example ends with a print out of the `classifier.train`. There is no explanation of what kinds of groups the light curves were classified into, and there are no graphs of light curves.

The docstrings are pretty spare and some of the attributes defined in `__init__` don't show up in the rendered API documentation because they are not `@property` so `automodapi` can't pick them up.

<table>
<tr>
<td width=150>Red</td>
<td>No documentation or some documentation, but very bare bones/minimal and incomplete or incorrect in a number of places.</td>
</tr>
</table>

### Testing

This area is "red". The overall test coverage number is 58%, but much of that comes from tests of their loading functions and utility functions. Only one of the classifier classes has tests, and there are no tests for autoencoder.

Running tests locally for version 0.1 gave two fails.  It looks like CI tests are failing in the repository
as of this writing.  In master there is still one fail.

Since tests are runnable by pytest, and they do have CI set up, it should be possible to get this to "orange" with some effort.  In particular better coverage of the main functionality of the package (based on the tutorial notebooks) would be needed.

<table>
<tr>
<td width=150>Red</td>
<td>No tests or tests that are not trivial to run or don’t use a standard testing framework, or low test coverage (no exact threshold for coverage since this is not always easy to measure, but in this category most of the code is not covered).</td>
</tr>
</table>

Test coverage can be tricky to measure, so this will be carefully assessed for
each package. The main idea is to determine whether it is low, medium or high
compared to what one might realistically achieve.

### Development status

There is a 0.1.0 release and development seems to be pretty active, so this is given "heavy development orange".

<table>
<tr>
<td>Heavy Development</td>
<td>Stable releases exist, but still under heavy development (so API changes can be frequent).</td>
</tr>
</table>

### Python 3 compatibility

This is "green" as it is Python 3 only.

<table>
<tr>
<td>Green</td>
<td>Compatible with Python 3</td>
</tr>
</table>

## Summary/Decision

Thanks for your work on this package! At the moment, we found a number of issues in the review areas. As per the review guidelines, we therefore won't be able to accept this package as an affiliated package yet. We will leave this pull request open for a month in case you would like to respond to the comments and/or address any of them.

If you have any follow-up questions or disagree with any of the comments above, leave a comment and we can discuss it here. At any point in future you can request a re-review of the package if you believe any of the scores should be updated - contact the coordination committee, and we’ll do a new review.

PR #327, opened 2019-06-06
Add coordinated packages to the affiliated page

eteq:
This PR updates the astropy web site to explicitly mention the concept of coordinated packages and highlight them in the package table as a separate section.  You can see this as it stands at http://eteq.github.io/astropy.github.com 

Three potential discussion items remain:

- @astrofrog, @taldcroft, and/or @kelle may want to comment on the text description of the coordinated packages.  I tried to channel both our discussions and the discussions at the Astropy coordination meetings, but I might have left something out...
- I'm not sure we ever agreed on this, but I decided to add the core package to the registry as a coordinated package. Debatably we should have a "review" of that package by someone who's not a core contributor (or at least not a coordinator), but at least for now I put it in as all "good"s.
- As it stands right now, the coordinated package maintainer names appear *twice* - once in the package registry, and a second time on the roles page.  It would be better to do this by reference somehow... But the difficulty is that 1) the "roles" page distinguishes lead and deputy while the registry does not, and 2) the registry is the "official" list, so we need the names there. There are solutions that could work, but it would take a fair amount of extra time to make the js changes, so I'm thinking we get this in now and make an issue about that (and just keep them manually synced in the interim)


astrofrog:
A few thoughts:

* Now that the page talks about both coordinated and affiliated packages, maybe we should just rename the page to 'packages' and then the first intro paragraph could talk about 1) core, 2) coordinated, and 3) affiliated?

* There is too much vertical whitespace in the coordinates packages table (and the affiliated table actually)

* The description of coordinated packages looks good

* I'm on the fence about having the core package in the table

* Agree about punting on the duplicate maintainer names

bsipocz:
giles hasn't picked up the latest commits, right?

astrofrog:
@bsipocz - I think it has?

eteq:
(just sent one more up, which I think makes it now all consistent)

bsipocz:
not it has (previously it was the list of 9 with regions still missing but astroplan  listed)

bsipocz:
If we rename the page to "packages" then having the core package there totally makes sense. Maybe it can have its own section, or can be left in the coordinated list?

bsipocz:
but for maintainers for the core package, please link to the team page to the list of maintainers instead just saying the coordinators.

eteq:
@bsipocz  re: 
> maintainers for the core package, please link to the team page 

I agree in principle, but unfortunately the javascripts functions fail in not-obvious-how-to-fix ways if I put links in there.  I'm sure it's fixable, but see my last bullet point in the OP about "lets get something in now and improve it later with better linking"?

bsipocz:
then describe it differently without the link. E.g. "Astropy maintainers, for details see the Team page above"

eteq:
@astrofrog:
> Now that the page talks about both coordinated and affiliated packages, maybe we should just rename the page to 'packages' and then the first intro paragraph could talk about 1) core, 2) coordinated, and 3) affiliated?


I see your point about renaming to "packages".  But that raises a couple new questions:

1. Should we rename it in the header bar on all the other pages?  If so, it might be a bit confusing for the header to just say "packages" without other context on the other pages?  Or maybe I'm worrying too much?
2. We really then should re-name the *directory* too, which changes the path to the json file... which we've in the past viewed as a "permalink" (although I'm not sure if anything actually uses it...?). Are we ok with that? Ideally this would be solved via a redirect, but I'm not sure if github pages supports that at the server level?


> There is too much vertical whitespace in the coordinates packages table (and the affiliated table actually)

I agree, but I cannot for the life of me figure out which CSS property to change to get rid of the whitespace.  I am open to ideas!

> I'm on the fence about having the core package in the table

Do you like it better if it's "packages" and with the changes brigitta suggested? I think it's not that unreasonable to consider the core to *be* a coordinated package.  It's just extra-well-coordinated...

astrofrog:
> Should we rename it in the header bar on all the other pages? If so, it might be a bit confusing for the header to just say "packages" without other context on the other pages? Or maybe I'm worrying too much?

I think all pages should have a consistent header - I personally don't find 'Packages' too ambiguous.

And yes I'm ok with having the core package in the list.

astrofrog:
See https://github.com/astropy/astropy.github.com/pull/331 for the vertical spacing

taldcroft:
FYI I'm hoping to look at this tomorrow (astropy Friday).

taldcroft:
Sorry for late input.. 

I'd suggest a slightly different wording to start the "Coordinated packages" to define the concept:

*A subset of affiliated packages are considered to be "Astropy coordinated packages" and are maintained directly by the Astropy project.  In practice this means the Astropy coordination committee has administrative control  ...*

So this means that coordinated packages are not *different* from affiliated packages, just a specialized subset (a subclass if you will).  That was always my understanding.  If we go in this direction then it means taking astropy core back out since it is by definition not an affiliated package.

I think overall it will be less confusing, and more consistent with historical precedent, to keep core as a separate thing (the one ring to bind them) instead of just another coordinated package.  That's the way it is presented in https://www.astropy.org/about.html, or the opening paragraph of https://astropy.readthedocs.io/en/latest/index.html, and the 2 papers and most talks about the Astropy project: core + affiliated packages.

With this, the top-level tab on astropy.org would stay as `Affiliated packages`.  That again maintains consistency with the historical precedent (since basically day 1) of the project.

astrofrog:
I agree with @taldcroft 👍

bsipocz:
Where do infrastructure packages belong? Are they coordinated or another category? (E.g. if coordinated is a subset of the affiliated packages, then they cannot be coordinated, unless coordinated is a union of infrastructure+some affiliated packages).

taldcroft:
It does seem like infrastructure packages meet the functional definition of coordinated.

Can the infrastructure packages be included as affiliated packages?  One might question "astronomy-related" (e.g. `pytest-helpers`?), but otherwise they fit in.  Maybe we could make that liberal enough to include them?  After all they form the basis for the whole affiliated package ecosystem, so that is pretty astronomy related even if they can be used otherwise.

*An affiliated package is an astronomy-related Python package that is not part of the astropy core package, but has requested to be included as part of the Astropy Project community. These packages demonstrate a commitment to Astropy’s goals of improving reuse, interoperability, and interface standards for Python astronomy and astrophysics packages.*

bsipocz:
I don't think they fit the criteria of being an "affiliated" package, but we can still include them in the definition of the coordinated packages. After all they are coordinated by the project, and being coordinated, in my view, is nothing more than a minimal guarantee that those packages are maintained by people close to the project, and are closely developed to the core package.

bsipocz:
This is what I have in mind: 
astropy package ecosystem(2)

astrofrog:
Side note - I think we should tweak the definition:

*An affiliated package is an astronomy-related Python package that is not part of the astropy core package, but has requested to be included as part of the Astropy Project community. These packages demonstrate a commitment to Astropy’s goals of improving reuse, interoperability, and interface standards for Python astronomy and astrophysics packages.*

to read

*An affiliated package is an astronomy-related Python package that is not part of the astropy core package, but **is a part of** the Astropy Project community. These packages demonstrate a commitment to Astropy’s goals of improving reuse, interoperability, and interface standards for Python astronomy and astrophysics packages.*

Otherwise requesting is enough to be affiliated.

In any case, I think it would be cleaner if infrastructure packages had their own table, so then we don't need to worry about the semantics of whether they are affiliated/coordinated/both.

I'd also like to suggest that we don't need to say anywhere whether coordinated packages are affiliated or not - we just say they are packages maintained by the project. We've used the term 'affiliated' so much that it's become a bit overloaded, but really when I think about it, out own coordinated packages aren't affiliated with us since they *are* developed by us - affiliated to me means third-party. So then I'd suggest we have:

**Coordinate packages**
(packages other than the core that we maintain, don't mention anything about affiliated or not)

**Infrastructure packages**
(all the infrastructure packages we maintain)

**Affiliated packages**
(these are third party packages that satisfy ...)

taldcroft:
I agree with @astrofrog.

bsipocz:
I like your suggestion, too, and totally agree that the affiliated word is overloaded, seen it being used casually by many people for repos that are using the template regardless whether they are indeed affiliated or not.

eteq:
(note: I rebase this and updated http://eteq.github.io/astropy.github.com/ to see how it looks post- #331)

eteq:
I think I also like @astrofrog's categorization here. I also see @taldcroft's argument in https://github.com/astropy/astropy.github.com/pull/327#issuecomment-513744413 about the core.

This still leaves us with the issue of having to rename the "Affiliated" header... But as a way to make incremental progress, I propose that I:
1. Update the *text* on the page to reflect https://github.com/astropy/astropy.github.com/pull/327#issuecomment-514199590
2. Merge this otherwise mostly as-is (with the core removed, though)
3. Add another issue to add the "infrastructure" packages formally (currently they're not in the list)
4. Leave the "affiliated" wording in the URLs and the nav bar as-is, but create an issue that's basically "we should fix this when we can work out the redirection logistics and such"

sound like a plan @taldcroft , @astrofrog , and @bsipocz ?

affiliated/index.html
```diff
@@ -73,13 +73,17 @@
...
+
+	<h1>Astropy Infrastructure Packages</h1>
+
+	<p>One final category is the "Astropy infrastructure packages". These packages are parts of the Astropy community's package infrastructure - e.g. testing and documentation machinery. In general they are more generic packages that are useful to packages outside of astronomy, and as a result may have some third-party maintainers. But some are also Astropy coordinates, generally falling under more generic infrastructure roles in the project. </p>
```
bsipocz: I would suggest to change "third party" to something else, maybe maintainers outside of the astropy project?

eteq: Good point - does the current wording look good?

bsipocz: yes, now looks good, except the typo

PR #331, opened 2019-06-27
Aesthetic fixes to website

astrofrog:
* Removed fontawesome CSS which I think is no longer needed
* Fixed vertical spacing in affiliated package table
* Fixed first column of affiliated package table which was too wide


astrofrog:
Preview at https://270-2081290-gh.circle-artifacts.com/0/html/affiliated/index.html

PR #335, opened 2019-07-29
add juan luis to coordinates

eteq:
Just what it says on the tin @Juanlu001


bsipocz:
oh, we do non ascii characters now? Can I have my accents, too? 😄

PR #339, opened 2019-09-10
Update GSoC deputies

eteq:
Adds @mirca as GSoC deputy, and remove @taldcroft (as he was there for the transition from last year).

cc @bsipocz

PR #340, opened 2019-09-10
fix role spellings (hooray unicode!)

eteq:
Given that I learned in #339 that roles.json is perfectly happy to accept non-ASCII characters, I realized some misspellings can be uniformly corrected.

cc @bsipocz as the primary recipient of the corrections...


bsipocz:
💯 
Thanks @eteq.

PR #355, opened 2019-11-05
Drop / change roles based on inputs from maintainer survey

taldcroft:


astrofrog:
@taldcroft - can you remove the 'Would prefer to be deputy' coloring and footnote, and keep @eteq as lead for the website and change @pllim to be deputy for votable?

taldcroft:
@astrofrog - I'm not 100% clear on the "keep @eteq as lead for the website".  This PR didn't change the website (documentation infrastructure) lead or deputy.  Do you want to swap Robitaille and Tollerud for the lead/deputy for the website?  Or am I completely misunderstanding?

astrofrog:
@taldcroft - this looks right, I just meant that the '1' should be removed from erik's name which you did 👍

bsipocz:
so this means we don't have anyone to maintain cosmology?

astrofrog:
@bsipocz - yeah there are some roles we definitely need to fill, for now we were removing people based on the survey but next we need to recruit people to fill roles! 😅

Issue #326, opened 2019-06-06
Re-arrange roles JSON to be more readable


eteq:
As brought up in https://github.com/astropy/astropy.github.com/pull/325#issuecomment-499260458 (and me internally every time I have to edit that section...), the roles.json file is very awkward to use for the sections with a lot of "sub-roles".  It would be better if instead of the form being:
```
...
"sub-role" : ["1", "2", "3"],
"lead" : ["a", "b", "c"],
"deputy" : ["d", "e", "f"],
...
```

instead it was something like 
```
...
"sub-role" : [{"name": "1", "lead": ["a"], deputy: ["d"] },
                     {"name": "2", "lead": ["b"], deputy: ["e"] },
                     {"name": "3", "lead": ["c"], deputy: ["f"] }]
...
```
that would make updating the list a lot easier to read.  But would require changes to the page javascript...

pllim:
I was just about to open an issue about this, and I am glad there is already one. :100:

PR #357, opened 2019-12-02
Clean up roles JSON

astrofrog:
This finally re-organizes the JSON file for the roles to be easier to read and update, in particular in relation to sub-roles. This makes the associated JS code simpler (in my opinion).

I've also removed @hamogu from the distribution role (it was for MacPorts) following his response in the roles survey, and fixed a missing description for the GSoC role that caused it to be absent from the roles descriptions.

[Preview](https://341-2081290-gh.circle-artifacts.com/0/html/team.html)

Fixes https://github.com/astropy/astropy.github.com/issues/326

### DIFF
--- a/validate_roles.py
+++ b/validate_roles.py
@@ -61,12 +61,12 @@ for i, role in enumerate(roles):
         if isinstance(role['responsibilities'], list):
             for resp in role['responsibilities']:
                 error += assert_string(i, 'responsibilities[description]', resp['description'])
-                for detail in resp['detail']:
+                for detail in resp['details']:
                     error += assert_string(i, 'responsibilities[detail]', detail)
         else:
             resp = role['responsibilities']
             error += assert_string(i, 'responsibilities[description]', resp['description'])
-            for detail in resp['detail']:
+            for detail in resp['details']:
                 error += assert_string(i, 'responsibilities[detail]', detail)
 
 if error > 0:
--- a/validate_roles.py
+++ b/validate_roles.py
@@ -3,15 +3,23 @@ import sys
 import json
 from termcolor import cprint
 
-REQUIRED_KEYS = {"role", "url", "sub-role", "lead", "deputy", "role-head",
-                 "role-subhead", "description", "sub-description"}
+REQUIRED_KEYS = {"role", "url", "role-head", "responsibilities"}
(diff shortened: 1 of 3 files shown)


taldcroft:
This looks awesome!  I can't find the rendered version though, I though the circleci test put that somewhere to look at (the artifacts?).

My only comment would be about making the role leads and deputies be a proper list of names instead of a comma-delimited string.  Eventually we'll phase out deputy and I think it's nicer as a data structure to have a list of names.

astrofrog:
@taldcroft - here is a preview: https://341-2081290-gh.circle-artifacts.com/0/html/team.html (giles seems to be down for some reason - @Cadair?)

I can change the comma-delimited fields to lists - I agree it would be a nicer structure.

astrofrog:
@taldcroft - here's the latest preview: https://342-2081290-gh.circle-artifacts.com/0/html/team.html

validate_roles.py
```diff
@@ -3,15 +3,23 @@
...
 
-REQUIRED_KEYS = {"role", "url", "sub-role", "lead", "deputy", "role-head",
-                 "role-subhead", "description", "sub-description"}
+REQUIRED_KEYS = {"role", "url", "role-head", "responsibilities"}
 
-LIST_KEYS = {"sub-role", "lead", "deputy", "role-head", "role-subhead",
-             "description", "sub-description"}
 
-STRING_KEYS = {"role", "url"}
+def assert_list(i, key, value):
```
eteq: (very minor)
```suggestion
def assert_is_list(i, key, value):
```
(would also require updating where used below)

validate_roles.py
```diff
@@ -3,15 +3,23 @@
...
+    if isinstance(value, list):
+        return 0
+    else:
+        cprint(f"   ERROR: Key \"{key}\" for role #{i} should be a list but instead is a {type(value)}", file=sys.stderr, color='red')
+        return 1
+
 
-LENGTH_MATCH_KEYSETS = [{"sub-role", "lead", "deputy"}]
+def assert_string(i, key, value):
```
eteq:
```suggestion
def assert_is_string(i, key, value):
```
(goes with the other suggestion)

PR #359, opened 2019-12-02
Add subsection to Contribute page about justifying effort

taldcroft:
The primary change is adding a section about justifying the time for contributing to Astropy in an academic setting.

Along the way I noticed that the first two sub-sections were `<h1>` when they really should be `<h2>` to fit in with the rest.  I also saw that the `href` for the Donate section seems wrong.

### DIFF
--- a/contribute.html
+++ b/contribute.html
@@ -77,7 +77,7 @@
 	</section>
 
 	<section id="feedback">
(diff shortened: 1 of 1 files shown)


eteq (approved):

eteq:
I'm happy with this as originally-written, and since there's been no other discussion I went forward with it.

PR #363, opened 2019-12-17
Updating version number to 4.0

bsipocz:
What with merging until the announcement is ready to go!

### DIFF
--- a/announcements/release-4.0.html
+++ b/announcements/release-4.0.html
@@ -114,7 +114,7 @@
 		<p>
 		&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<a href="http://docs.astropy.org/en/stable/whatsnew/4.0.html">http://docs.astropy.org/en/stable/whatsnew/4.0.html</a>
 		</p>
-		<p>Astropy v4.0 now replaces v2.0 as the long term support release, and will be supported until the end of 2021. 
+		<p>The Astropy v4.0.x series now replaces v2.0.x as the long term support release, and will be supported until the end of 2021. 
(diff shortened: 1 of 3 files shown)


bsipocz:
@eteq - note that I've reordered the contributor list by first name, as that one is unambiguous for ordering and makes life so much easier to auto generate from the what's new page via credit.rst than manually inserting the ~20-30 new ones at the right place at release time.

bsipocz:
> I might consider trying to figure out how to make this last-name-alpha, but better to just accept this and do that as a follow on if need be.

Folks may have various combinations of names. For a multiple entry list how would you figure out which one to sort one, whether one is a second, third, etc given name or the first last name, etc?

eteq:
@bsipocz - I added a release announcement in e6c7cf2d6262469e135df357012cbcc8596e1a3f - can you have a quick look at it?

announcements/release-4.0.html
```diff
@@ -0,0 +1,188 @@
...
+			<li>Improved downloading and caching of remote files</li>
+		</ul>
+		</p>
+		<p>
+		In addition, hundreds of smaller improvements and fixes have been made. An overview of the changes is provided at:
+		</p>
+		<p>
+		&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<a href="http://docs.astropy.org/en/stable/whatsnew/4.0.html">http://docs.astropy.org/en/stable/whatsnew/4.0.html</a>
+		</p>
+		<p>Astropy v4.0 now replaces v2.0 as the long term support release, and will be supported until the end of 2021.
```
bsipocz: Call these also the v4.0.x series or branch, etc, as 4.0 is replacing 2.0.16 rather than 2.0.

eteq:
Addressed @bsipocz comment - once tests pass I'll merge and we can send things out!

PR #367, opened 2020-01-21
Fix Giles URL

pllim:
cc @Cadair


bsipocz:
OK, Giles came up as a check as expected.

Thanks @pllim!

PR #369, opened 2020-03-13
Update mocpy web site

eteq:
mocpy's docs to have been updated to use the github pages URL instead of an RTD url.  This just updates that to get the CI back and working for this repo.

@bmatthieu3 can probably confirm that this is the right URL - please approve or :+1: if you agree with this


bsipocz:
👍 We already did this for astroquery in https://github.com/astropy/astroquery/pull/1639

Issue #364, opened 2019-12-17
Update roles list to combine Lead/Deputy


eteq:
As discussed and unanimously agreed to at the Astropy Coordination Meeting, we are combining the "lead" and "deputy" columns in the roles list.

This is conceptually straightforward but requires some re-work of the javascript...

PR #376, opened 2020-05-10
Added affiliated package editorial role

astrofrog:
This adds the affiliated package editorial role to the list of roles

cc @hamogu @pllim (can't assign you for review due to team permissions for now but clearly we will need to add permissions for you for this repo)

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -169,7 +169,7 @@
     {
         "role": "Affiliated package review editor",
         "url": "Affiliated_package_review_editor",
-        "lead": ["Pey Lian Lim", "Hans Moritz G\u00fcnther"],
+        "lead": ["Hans Moritz G\u00fcnther", "Pey Lian Lim"],
         "deputy": ["Unfilled"],
         "role-head": "Affiliated package review editor",
         "responsibilities": {
--- a/roles.json
+++ b/roles.json
@@ -177,7 +177,8 @@
             "details": [
                 "Keeping track of submissions to the affiliated package ecosystem",
                 "Managing the review process for packages",
-                "Making decisions regarding accepting and rejecting packages based on reviews"
+                "Making decisions regarding accepting and rejecting packages based on reviews",
+                "Organizing the re-review of packages that have not been recently reviewed"
             ]
         }
     },
--- a/roles.json
+++ b/roles.json
@@ -156,7 +156,7 @@
     {
         "role": "Workshops coordinator",
         "url": "Astropy_workshops_coordinator",
-        "lead": ["David Shupe, Brett Morris"],
+        "lead": ["David Shupe", "Brett Morris"],
         "deputy": ["Unfilled"],
         "role-head": "Astropy Workshops coordinator",
         "responsibilities": {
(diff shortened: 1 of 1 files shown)


roles.json
```diff
@@ -166,6 +166,22 @@
...
         }
     },
+    {
+        "role": "Affiliated package review editor",
+        "url": "Affiliated_package_review_editor",
+        "lead": ["Pey-Lian Lim", "Hans Moritz G\u00fcnther"],
+        "deputy": ["Unfilled"],
+        "role-head": "Affiliated package review editor",
+        "responsibilities": {
+            "description": "Oversee the affiliated package submission and review process, including:",
```
hamogu, line 176: is it wise end on ", including"? What if the change the display format in the future to a version where the details are not shown by default or are shown in another location?

hamogu: I looked the rest of the code and I realize that that's how it works and my comment is stupid - no change required.

roles.json
```diff
@@ -166,6 +166,22 @@
                         "Identify opportunities for workshops in diverse geographic locations"]
         }
     },
+    {
+        "role": "Affiliated package review editor",
+        "url": "Affiliated_package_review_editor",
+        "lead": ["Pey-Lian Lim", "Hans Moritz G\u00fcnther"],
+        "deputy": ["Unfilled"],
```
pllim, line 173: I thought we decided to do away with lead/deputy at 2019 meeting, no? Or does this mean we need one more to act as deputy?

roles.json
```diff
@@ -166,6 +166,22 @@
                         "Identify opportunities for workshops in diverse geographic locations"]
         }
     },
+    {
+        "role": "Affiliated package review editor",
+        "url": "Affiliated_package_review_editor",
+        "lead": ["Pey-Lian Lim", "Hans Moritz G\u00fcnther"],
```
pllim, line 172: Hyphen is not official.
```suggestion
        "lead": ["Pey Lian Lim", "Hans Moritz G\u00fcnther"],
```

Also, how are these names ordered? Either by first or last name, @hamogu should have been listed first.

hamogu: I saw that, too, and concluded that it's probably reverse alphabetically ordered (either by first of by last name, could not figure that out)  ;-)

roles.json
astrofrog: Yes we did decide that but the webpage still needs to be updated in a separate PR

astrofrog: See https://github.com/astropy/astropy.github.com/issues/364

roles.json
```diff
@@ -166,6 +166,22 @@
                         "Identify opportunities for workshops in diverse geographic locations"]
         }
     },
+    {
+        "role": "Affiliated package review editor",
+        "url": "Affiliated_package_review_editor",
+        "lead": ["Pey Lian Lim", "Hans Moritz G\u00fcnther"],
```
pllim, line 172: What about the ordering?

roles.json
astrofrog: it was in random order ;) I just re-ran with a new random seed and got alphabetical, so I'll go with that!

roles.json
astrofrog: Fixed, see below

astrofrog:
@kelle and @taldcroft approved verbally, so merging!

PR #383, opened 2020-06-22
Update code_of_conduct.html to include political views in the protected category

logankilpatrick:
Consider adding a note on non-discrimination based on political views.

We at the Julia Language just ran into an issue related to this and updated our code of conduct.

### DIFF
--- a/code_of_conduct.html
+++ b/code_of_conduct.html
@@ -91,7 +91,7 @@
 		<li>We pledge to treat all people with respect and provide a harassment- and
 		bullying-free environment, regardless of sex, sexual orientation and/or
 		gender identity, disability, physical appearance, body size, race,
-		nationality, ethnicity, and religion. In particular, sexual language and
+		nationality, ethnicity, political view, and religion. In particular, sexual language and
 		imagery, sexist, racist, or otherwise exclusionary jokes are not
 		appropriate.</li>
 


taldcroft:
If we're being explicit, it seems that in general politics should not be part of any astropy conversation except in the narrow context of directly impacting astropy (funding, for instance). In our community there is a strongly under-represented political view, and I expect that any members that hold the minority view feel excluded by typical casual political conversation, e.g. at a bar during a conference.

astrofrog:
I think we need to be careful here because certain political views come with intolerance to some of the other categories, so I think we might want to make explicit that the tolerance to the political views cannot he used as an excuse for intolerance to the other categories (to some extent the same is true of religious views which might translate into intolerance of other categories). Not sure what the best way is to convey this though.

taldcroft:
@astrofrog - that's kind of why I lean toward an approach of discouraging discussion of politics in general, sort of like how American families keep the Thanksgiving dinner pleasant. It's all pretty tricky. Note that many people would view the banner on the astropy.org page as an overtly political statement, so it is all a matter of perspective.

hamogu:
I think the wording outlaws behaviour that falls under any one of those categories. If my religion requires me to hate all Vikings, harassing Danish and Swedish contributors to Astropy remains forbidden by the code of conduct, even though my religion per-se is protected. Similarly, if my political views are expressed in sexist or discriminatory language, they still violate the code of conduct. 

What this does is, that it tell us that we do not want to give preference to any particular political view, the same way that we do not want to to give preference to rich, white guys only.

pllim:
So, do we have a consensus of a way forward here? I don't think I see one.

logankilpatrick:
Agreed, we can close this. Thanks all for the input.

pllim:
Nevertheless, thanks for your PR, @logankilpatrick !

PR #384, opened 2020-06-22
Update index.html to include note on pip3 install

logankilpatrick:
Add note on pip3 install in case someone is like me and has python 2.7 and 3.x.


pllim:
I am not sure if this is relevant because Astropy project has dropped Python 2 support since v3.0. And Python 2 has reached EOL back in Jan 2020, so people should not be using it anymore. Thanks for your contribution anyway!

PR #398, opened 2020-10-01
Remove @eteq from docs and testing infrastructure roles

astrofrog:
As discussed by email with @eteq (since these roles are well populated)


pllim:
😱

eteq (approved):
To be clear: I am still interested in these things, but I'm trying to not be a roadblock since I'm clearly over-committed right now.

PR #399, opened 2020-10-07
Affiliated packages now must be 3.x compatible.

### LINKED ISSUE #233: "Becoming an Affiliated Package" suggests Python 3 is still optional
The instructions on "[Becoming an Affiliated Package](http://www.astropy.org/affiliated/)" talk about Python 3 compatibility being "encouraged", and that compatibility via the _six_ package is recommended. However, both of these statements are in disagreement with the linked [Coding Guidelines](http://docs.astropy.org/en/stable/development/codeguide.html#writing-portable-code-for-python-2-and-3) - Python 3 is mandatory, and _six_, etc., is not permitted.

hamogu:
fixes #233

### DIFF
--- a/affiliated/index.html
+++ b/affiliated/index.html
@@ -244,12 +244,7 @@
 		possible, but when it is, a test suite is a crucial element of stable
 		software and reproducible science.</li>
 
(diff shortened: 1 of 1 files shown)


pllim (approved):
LGTM. Thanks!

PR #400, opened 2020-10-09
Add new DevOps role

astrofrog:
I'd like to suggest we move ahead with creating a 'DevOps' role that would initially be populated with @embray, @pllim and @bsipocz. I've therefore added such a role to the list here, but **mainly with placeholder information**. I'm not thrilled about the current role title, and the list of responsibilities is just a placeholder.

This role would be a significant one as it would come with admin access to all repositories in the org to enable some of the responsibilities I've listed (although we could investigate whether maintainer access is sufficient - basically whatever level is needed to make it possible to do the things).

At this point, what I'd like is for @embray @pllim @bsipocz to take a look at this and make concrete suggestions on what you'd like the role title and responsibilities to be - feel free to either make GitHub suggestions, or a PR against my branch, or commit directly to my branch if you can. Once you are happy with the PR, we could then put it to the CoCo for approval (I'd rather this approach than having the role be defined by the CoCo since @bsipocz and @pllim in particular you know best what you've been doing in this currently unofficial role and what you'd like to be able to do but not been able to for lack of permissions).

Once this is merged in, we should email astropy-dev so that people are aware of this new role.

xref https://github.com/astropy/astropy-project/issues/118 https://github.com/astropy/astropy-project/issues/136 https://github.com/astropy/astropy-project/issues/149

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -281,8 +281,8 @@
             "details": [
                 "Set up and maintain continuous integration services",
                 "Ensure adequate labeling of issues and pull requests",
-                "Perform triaging of issues and pull requests, including moving between repositories",
-                "Merging non-controversial pull requests"
+                "Perform initial triaging of issues and pull requests, including moving between repositories",
+                "Merge non-controversial pull requests"
             ]
         }
     },
--- a/roles.json
+++ b/roles.json
@@ -268,6 +268,25 @@
         }
     },
     {
+        "role": "DevOps and Operations",
+        "url": "devops_team",
+        "people": [
+            "Madison Bray",
+            "Pey Lian Lim",
+            "Brigitta Sip\u0151cz"
+        ],
+        "role-head": "DevOps and Operations",
+        "responsibilities": {
+            "description": "Ensure the smooth running of the project",
+            "details": [
+                "Set up and maintain continuous integration services",
+                "Ensure adequate labelling of issues and pull requests",
+                "Perform triaging of issues and pull requests, including moving between repositories",
+                "Merging non-controversial pull requests"
+            ]
+        }
+    },
+    {
         "role": "Testing infrastructure maintainer",
         "url": "Testing_infrastructure_maintainer",
         "people": [
--- a/roles.json
+++ b/roles.json
@@ -271,7 +271,7 @@
         "role": "DevOps and Operations",
         "url": "devops_team",
         "people": [
-            "Madison Bray",
+            "E. Madison Bray",
             "Pey Lian Lim",
             "Brigitta Sip\u0151cz"
         ],
--- a/roles.json
+++ b/roles.json
@@ -268,7 +268,7 @@
         }
     },
     {
-        "role": "DevOps and Operations",
+        "role": "DevOps and Operations Specialist",
         "url": "devops_team",
         "people": [
             "E. Madison Bray",
--- a/roles.json
+++ b/roles.json
@@ -280,7 +280,7 @@
             "description": "Ensure the smooth running of the project",
             "details": [
                 "Set up and maintain continuous integration services",
-                "Ensure adequate labelling of issues and pull requests",
+                "Ensure adequate labeling of issues and pull requests",
                 "Perform triaging of issues and pull requests, including moving between repositories",
                 "Merging non-controversial pull requests"
             ]


taldcroft (approved):
:+1:

roles.json
```diff
@@ -267,6 +267,25 @@
             ]
         }
     },
+    {
+        "role": "DevOps and Operations",
+        "url": "devops_team",
+        "people": [
+            "Madison Bray",
```
pllim, line 274: I think this is more correct? But let's see what @embray says.
```suggestion
            "E. Madison Bray",
```

roles.json
```diff
@@ -267,6 +267,25 @@
...
+            "Pey Lian Lim",
+            "Brigitta Sip\u0151cz"
+        ],
+        "role-head": "DevOps and Operations",
+        "responsibilities": {
+            "description": "Ensure the smooth running of the project",
+            "details": [
+                "Set up and maintain continuous integration services",
+                "Ensure adequate labelling of issues and pull requests",
```
pllim, line 283: If we are doing US English, it's only one "l". ⚔️ 
```suggestion
                "Ensure adequate labeling of issues and pull requests",
```

pllim (approved):
Just minor comments. I think current descriptions are good for a start. We can always improve it as subsequent PRs.

As defined in Wikipedia, [DevOps](https://en.wikipedia.org/wiki/DevOps) also includes "releasing." While the role here helps with Continuous Delivery, it does not perform the "manual release" part of CD. So if you want to be pedantic, it is really "most of DevOps." 😹

roles.json
```diff
@@ -267,6 +267,25 @@
             ]
         }
     },
+    {
+        "role": "DevOps and Operations",
```
eteq, line 271:
```suggestion
        "role": "DevOps and Issue Triage Maintainer",
```
I'm not a fan of "DevOps and Operations" because it's operations twice.  But maybe this captures it better?  Not that I'm thinking "issue triage maintainer" on the theory that this is not necessarily the people who do all the triaging (even though in practice right now, it's definitely the lion's share!) but potentially it can be the people who oversee a larger group doing triaging.

eteq (commented):
Yep, I like this role broadly speaking, but I offered an alternative name suggestion inline.

Also, a broader question: should this subsume some of the existing roles that overlap with it?  Particularly I'm thinking about "CI-helpers maintainer" and maybe "Testing infrastructure maintainer", both of which are pretty deeply intertwined with the CI and therefore DevOps.

roles.json
astrofrog: I think this is perhaps too specific.

pllim: In reality it is "repo janitor" but it doesn't sound nice. 😉

bsipocz: Indeed, this is the role that goes into the gutter of things :)

issue triage is something different though, that should very much include the contextual package maintainers and that is need to be done regularly. devops people cannot really go into the issues and triage them besides an initial classification (and frankly that part is the easiest entry-level thing that can be outsourced to brand new people to I would not tie it with admin access). 

tbh I quite like "devops and operations" as that is something the industry knows and understand, coming up a flowery language to describe it on the other hand would not have the same effect.

bsipocz:
subsume ci-helpers is 👍 as that is exactly about operations, but the plugins I think are more than patching up the actual fires (though quite often it is just that that). So the former has to have the devops people involved, the latter, especially new fucntionality development can certainly be overseen by others.

roles.json
eteq: Ok I see these points.  My main reason for adding "issue triage maintainer" is because the roles specifically say "Perform triaging of issues and pull requests" and that is *not* normally a DevOps role.  But DevOps is not well definied in s/w engineering IMHO anyway so :shrug:.

But as a compromise, what about just
```suggestion
        "role": "DevOps Maintainer",
```
 then? I still think "DevOps and Operations" is like "SDSS Survey".

(And even if you don't like that, I think some other re-wording is necessary for it to be a *person in a role* rather than a role.  That is, to be consistent with the other roles it should be "DevOps maintainer" "DevOps specialist", "DevOps Ninja" or similar, not just "DevOps".)

eteq:
FYI I went ahead and created a github team for this role with admin access to astropy/astropy.  We can add other repos at request, although I presume by default it makes sense for all coordinate/infrastructure packages to be added?

Either way, once this is finalized as a role I will add the people in it to the github team and we can use that in the future to give the relevant repo-level admin rights.

roles.json
```diff
@@ -267,6 +267,25 @@
...
+            "Pey Lian Lim",
+            "Brigitta Sip\u0151cz"
+        ],
+        "role-head": "DevOps and Operations",
+        "responsibilities": {
+            "description": "Ensure the smooth running of the project",
+            "details": [
+                "Set up and maintain continuous integration services",
+                "Ensure adequate labeling of issues and pull requests",
+                "Perform triaging of issues and pull requests, including moving between repositories",
```
bsipocz, line 284: It should be initial labeling and triaging, as ongoing triage should involve the subpackage maintainers

roles.json
bsipocz: it's not really maintainer, as part of the role is to get things fixed up, worked around, etc, not just fixing up the tools that are needed for things to get fixed up. Thus my liking for devops and operations, even though it's RAS (but if it works for sdss, it could work for this, too :) )
While it somewhat sounds weird, could live with the "devops specialist", too, but I feel "ninja" is very much curturaly inappropriate.

roles.json
astrofrog: Sounds good, will update

astrofrog:
I've made some tweaks based on the comments above but left the name of the role the same for now. Since better is the enemy of good, any objections to going with the current name and description and updating via PRs if needed later? Regarding the role naming, I think at the end of the day the final decision rests on the people in the role 😆

bsipocz (approved):

roles.json
eteq: Ok, I'm cool with either
```suggestion
        "role": "DevOps Specialist",
```
or 
```suggestion
        "role": "DevOps and Operations Specialist",
```
then.  I'm open to really any "person" ending there too.

astrofrog:
I tweaked the role title for grammatical consistency with other roles

astrofrog:
@bsipocz @pllim @embray - feel free to open follow-up PRs if you want to refine the role description over time! :)

PR #405, opened 2020-10-14
Update current version to v4.0.3

astrofrog:
v4.0.3 is now released!


pllim (approved):

pllim:
Congrats! I am wondering if we could make that element programmatic but maybe for future hack days!

astrofrog:
Yes we could just use the PyPI JSON API to dynamically get this info on the homepage

saimn:
An easy solution would be to use a badge:

- https://badge.fury.io/for/py/astropy
[PyPI version](https://badge.fury.io/py/astropy)

- https://img.shields.io/pypi/v/astropy
[PyPI version](https://img.shields.io/pypi/v/astropy)

PR #413, opened 2020-11-09
Change gala pypi name

adrn:
I recently got access to the name "gala", bringing the pypi name and the github repo name in line, so I will be using this name going forward (🎉).

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -505,7 +505,7 @@
             "stable": true,
             "home_url": "http://gala.adrian.pw/en/latest/",
             "repo_url": "https://github.com/adrn/gala",
-            "pypi_name": "astro-gala",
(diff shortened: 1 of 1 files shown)


pllim:
Congrats! Do we need a "was astro-gala" in the description or do you think this is enough?

pllim:
Also what will happen to astro-gala on PyPI? Any chance another package can re-use that name? If so, we should put a warning.

adrn:
I'm going to keep astro-gala on pypi (for older versions) until someone asks for the name :) 

Good idea to add a note to the description - will do that in the next commit.

pllim (approved):
I would recommend you keep the old name forever, lest it be exploited. But if you release the name, please remember to come back and add a warning. Thanks!

adrn:
Thanks!

PR #415, opened 2020-11-30
DOC: Replace astropy channel with conda-forge

pllim:
Similar to astropy/astropy#11111

cc @mwcraig

### DIFF
--- a/announcements/release-4.1.html
+++ b/announcements/release-4.1.html
@@ -136,10 +136,10 @@ pip install astropy --upgrade
 conda update astropy
 </pre>
 		<p>
-		Or if you cannot wait for Anaconda to update their default version, you can use the astropy channel:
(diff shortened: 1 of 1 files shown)


pllim:
I thought we agreed on this, so I don't know what the hold up is.

mwcraig:
Is there a need to merge this if we merge astropy/astropy#11111? 

It doesn't seem like the 4.1 release announcement is the place that needs the fix given how long ago that was....

pllim:
Oh yah... Let's just close this then. When we finally really want to do this, me or some other person can open a new PR.

pllim:
Looks like last release announcement already uses conda-forge, so maybe we don't need this:

https://github.com/astropy/astropy.github.com/blob/2b68f19840ac89bcdb179adf64201db55f868ba3/announcements/release-4.3.html#L143

PR #416, opened 2020-12-01
Add section to Contribute page for Astropy role

taldcroft:
This adds a new section which describes assuming a formal Astropy Role as a way to contribute to Astropy.

PR #417, opened 2020-12-02
Contribute page cleanup & ToC

eteq:
Thinking of #416 as the "straw that broke the camel's back", I noticed while reviewing a few consistency issues and recognized that the page has gotten big enough that it could use a sort of table of contents.  So this PR does a general cleanup of the contribute page and adds some quick links to the top.

(You can see it live at http://eteq.github.io/astropy.github.com/contribute.html at the time of this writing)


eteq:
Oh and sorry if the review is a bit annoying - the last commit is a whitespace cleanup.  But the two other commits can be examined individually for the "real" content changes.

taldcroft (approved):
Looks good to me.  No worries about whitespace, the GitHub diff preferences allow ignoring it, which comes in handy quite a lot!

Issue #294, opened 2019-01-13
Bibtex copy to clipboard is missing backslashes for journal shortcuts


adrn:
As reported by https://twitter.com/mpoessel/status/1084476039185793025, for example: "aj" instead of "\aj".

bsipocz:
Note: the links to the bibtex provides the right format, as well as `astropy.__citation__`, the bug only comes up in the  "copy to clipboard" javascript...

PR #423, opened 2021-02-21
Escape `\`s in BibTeX

jdswinbank:
The BibTeX generated by the big orange button at https://www.astropy.org/acknowledging.html currently doesn't work, because all the `\`s are swallowed by Javascript. This escapes them, so they aren't.


jdswinbank:
See also #294 !

PR #406, opened 2020-10-15
Use PyPI badge for current release version

pllim:
Follow up of #405 . Thanks to @saimn for the idea.

PR #427, opened 2021-04-12
update for 4.2.1

eteq:
a two-character PR!


pllim (approved):
p.s. Maybe at some point, we should revisit #406

PR #448, opened 2021-10-14
Drop "interim" from Finance Committee name

kelle:


hamogu:
We generally have a two week period for comments when people are added to roles, but I don't think we need that when people resign from a role, since this is not really something that needs community consensus - if Kelle resigns from this role, that's her decision. So, merging.

PR #450, opened 2021-10-21
fix links to point to newer versions of affiliate instructions

eteq:
just what it says on the tin... just a couple of out-of-date links.


pllim (approved):
LGTM. Thanks!

PR #454, opened 2021-11-29
Remove link to “documentation overview”

jdswinbank:
That page says “This page has been removed. For an overview, see the
http://www.astropy.org page”, and has done since at least v2.0.16
(https://docs.astropy.org/en/v2.0.16/overview.html).


pllim (approved):
Good catch!

PR #455, opened 2021-12-03
Replace Moritz with Aarya on the finance committee

hamogu:
Moritz has said in this nomination for the Coco, that he would step back
from other Astropy roles if elected. Based on nominations from the community,
the Coco has contacted Aarya, who is interested in joining the finance
committee.
This PR will stay open for two weeks for any discussion.

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -129,7 +129,7 @@
         "url": "finance_committee_member",
         "people": [
             "Kelle Cruz",
-            "Hans Moritz G\u00fcnther",
+            "Aarya Patil",
             "John Swinbank",
             "Erik Tollerud"
         ],


jdswinbank:
Moritz has been a fantastic member of the Finance Ctte and I don't want him to leave. 😭 

But that said, I'm confident that Aarya will be equally fantastic, so I have no hesitation in supporting this proposal.

hamogu:
I'll merge this now, which adds @aaryapatil  to the finance committee and removes me. We have decided that a few weeks of overlap would be good for a smooth transition, so I'm going to attend a little longer, but, as of now, I'm no longer a formal member.

PR #464, opened 2022-01-19
Update release coordinator to release team and update membership

astrofrog:
@saimn @embray @bsipocz - could you check over the responsibilities and suggest changes/additions? The bottom line is I think the responsibility of the team should be the core package releases and then oversight of the release process for other organization packages (but not necessarily doing those releases)

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -387,7 +387,7 @@
                 "Notifying the Distribution Coordinators of any core astropy package release",
                 "Working with the Community Engagement Coordinator to make release announcements via channels such as mailing lists and social media",
                 "Keeping documentation for the release process of the core package up to date",
-                "Designing policies to improve the uniformity of release procedures across the astropy organization (including coordinated and infrastructure packages)",
+                "Designing policies to improve the uniformity of release procedures for coordinated and infrastructure packages",
                 "Testing the interoperability of the core, coordinated, and infrastructure packages and provide safeguards against breaking the ecosystem."
             ]
         }
--- a/roles.json
+++ b/roles.json
@@ -387,7 +387,7 @@
                 "Notifying the Distribution Coordinators of any core astropy package release",
                 "Working with the Community Engagement Coordinator to make release announcements via channels such as mailing lists and social media",
                 "Keeping documentation for the release process of the core package up to date",
-                "Designing policies to improve the uniformity of release procedures for coordinated and infrastructure packages",
+                "Designing policies to improve the uniformity of release procedures for the coordinated and infrastructure packages of the Astropy Project",
                 "Testing the interoperability of the core, coordinated, and infrastructure packages and provide safeguards against breaking the ecosystem."
             ]
         }
--- a/roles.json
+++ b/roles.json
@@ -371,26 +371,24 @@
         }
     },
     {
-        "role": "Core package release coordinator",
-        "url": "Core_package_release_coordinator",
+        "role": "Release team",
+        "url": "release_team",
         "people": [
+            "E. Madison Bray",
+            "Simon Conseil",
             "Thomas Robitaille",
-            "Brigitta Sip\u0151cz",
-            "Erik Tollerud"
+            "Brigitta Sip\u0151cz"
         ],
-        "role-head": "Core package release coordinator",
+        "role-head": "Release team",
         "responsibilities": {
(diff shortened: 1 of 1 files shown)


bsipocz:
There are a lot of other packages/repos in the org, since are affiliated but since are but even affiliated with no interaction with the asteroid dev community. Those, I suggest, are not our responsibility and best would be to remove them to a separate outer org (see scipy vs scikit-, or the difference between the scikit-learn and sciit-learn-contrib org)

saimn:
Looks good to me, I don't have much to add to the list.

roles.json
```diff
@@ -371,26 +371,24 @@
...
-                "Maintaining the bugfix branches",
-                "Updating Astropy\u2019s PyPI entry as needed",
+                "Carrying out releases of the core astropy package",
+                "Notifying the Distribution Coordinators of any core astropy package release",
                 "Working with the Community Engagement Coordinator to make release announcements via channels such as mailing lists and social media",
-                "Notifying the Distribution Coordinators of any release",
-                "Make sure that the package gets updated in conda and coordinate with distribution coordinators",
-                "Keep documentation for release process up to date"
+                "Keeping documentation for the release process of the core package up to date",
+                "Designing policies to improve the uniformity of release procedures across the astropy organization (including coordinated and infrastructure packages)",
```
astrofrog, line 390: @bsipocz I will change this to say across infrastructure and coordinated packages, does that seem more reasonable? (we do need to tidy up the org but that is a separate issue)

bsipocz: yes, that's totally reasonable, thanks!

astrofrog: Done

roles.json
```diff
@@ -371,26 +371,24 @@
...
-                "Maintaining the bugfix branches",
-                "Updating Astropy\u2019s PyPI entry as needed",
+                "Carrying out releases of the core astropy package",
+                "Notifying the Distribution Coordinators of any core astropy package release",
                 "Working with the Community Engagement Coordinator to make release announcements via channels such as mailing lists and social media",
-                "Notifying the Distribution Coordinators of any release",
-                "Make sure that the package gets updated in conda and coordinate with distribution coordinators",
-                "Keep documentation for release process up to date"
+                "Keeping documentation for the release process of the core package up to date",
+                "Designing policies to improve the uniformity of release procedures for coordinated and infrastructure packages",
```
eteq, line 390:
```suggestion
                "Designing policies to improve the uniformity of release procedures for the Astropy Project, particularly coordinated and infrastructure packages",
```

eteq (commented):
One small tweak to suggest here: I think the reality is that some of the affiliated packages look to the core/coordinated/infrastructure packages for *guidance* and e.g. follow our release procedures so they don't have to make up their own. I think it's part of the remit to this team to at least consider that.  Totally makes sense that it should only be *responsible* for packages that are in the org (not necessarily even all of those, as @bsipocz points out), but it shouldn't be "this team can completely ignore affiliated packages", which the current wording sort of implies to me (and with the small suggestion I made, allows for it).

eteq:
Following APE0, this will need a two-week comment period since it's a change to the roles. I think we can consider that clock to have started when the PR was opened, so this can be merged Feb 2 assuming the CoCo agrees (which it has already on the general idea) and there are not community objections.

roles.json
bsipocz: 👎  on `particularly` as explicit is better than implicit. Adding the the Astropy Project to the end of the sentence would be good. `"for the coordinated and infrastructure packages of the Astropy Project" - so we are not taking up the maintenance of pieces of infra outside the project.

Of course, the team can be consulted for or asked to help out with affiliated packages, but this page describes the remit and responsibilities of this role. The project doesn't have mandate to with non coordinated packages, let alone design release procedure for downstream packages, if some of those follow ours, it's all good and nice, but given that there are ~50 affiliated packages, those should not fall inside our talks at all.

eteq: Ok, so is this what you're thinking @bsipocz ? I'm ok with that.
```suggestion
                "Designing policies to improve the uniformity of release procedures for the coordinated and infrastructure packages of the Astropy Project",
```

eteq:
The comment period has elapsed with no objections, so I'll merge this once @astrofrog has addressed (or rejected) the one remaining dangling thread above.

PR #467, opened 2022-01-27
Show latest core package version from PyPI

astrofrog:
This uses JS code adapted from

https://github.com/sunpy/sunpy-sphinx-theme/blob/main/sunpy_sphinx_theme/sunpy/static/version.js

to get the latest version from PyPI which will save having to update it during releases. This is an alternative to https://github.com/astropy/astropy.github.com/pull/406 and the advantage here is we maintain the same formatting as before.

@Cadair - do you want us to include the sunpy license or is this trivial enough to life without attribution? (either way is fine!)

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,6 +19,12 @@
 <!-- Meow woof -->
 <script>
 window.onload = function() {
+
+  $.getJSON('https://pypi.org/pypi/astropy/json', function(data) {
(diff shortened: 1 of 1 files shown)


Cadair:
nah it's fine, it's very trivial.

Cadair (approved):

adrn (approved):
Very sensible change -- thanks @astrofrog!!

PR #478, opened 2022-04-07
Fix maintainer name for pllim

pllim:
No dash. 😸 

- [x] Signed off by @pllim

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -178,7 +178,7 @@
         },
         {
             "name": "ginga",
-            "maintainer": "Eric Jeschke and Pey-Lian Lim <eric@naoj.org>",
+            "maintainer": "Eric Jeschke and Pey Lian Lim <eric@naoj.org>",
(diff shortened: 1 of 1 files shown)


hamogu:
While looking at that: Is that the single email for both maintainers? Or is that the email for the first maintainer and, if so, should it appear after the first name?

dhomeier:
> While looking at that: Is that the single email for both maintainers? Or is that the email for the first maintainer and, if so, should it appear after the first name?

1. no
2. yes
3. don't know. The first maintainer is also hosting the websites, so maybe that's the generic address for the package as well?

Besides the point, but the homepage also has a broken link to the developer manual (correctly at https://ginga.readthedocs.io/en/stable/dev_manual/).

pllim:
Re: https://github.com/astropy/astropy.github.com/pull/478#issuecomment-1091919818 -- What email? I am very confused. Just trying to fix my own name here...

PR #477, opened 2022-04-07
Update mocpy maintainer list

pllim:
This is to mirrror a change at astropy/astropy-v5.0-paper#81 . I think we should formalize it here first before putting that on the paper.

- [x] Signed off by @fxpineau
- [x] Signed off by @bmatthieu3

mocpy was added in #366

PR #482, opened 2022-04-07
Use Oxford comma for Einsteinpy

pllim:
Oxford comma! Should be uncontroversial? 

Actually, do we need to order by last name alphabetically?


dhomeier:
See https://github.com/astropy/astropy.github.com/pull/477#discussion_r845396565 – if the maintainers have something like a first-author order, we should not automatically override it.

PR #483, opened 2022-04-07
Update pyvo maintainers

pllim:
Added Tom D but also found a typo in Adrian D's name.

- [ ] Signed off by @tomdonaldson
- [ ] Signed off by @andamian


eteq:
This hasn't been approved here from @tomdonaldson or @andamian, but there was some "signing-off" on slack, so I think this is fine.  Merging!

PR #486, opened 2022-04-21
Remove "4 people" from coco role definition

adrn:
The number is wrong, and it is documented in the APE, so this is probably TMI for the team page.

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -9,7 +9,7 @@
             "Adrian Price-Whelan",
             "Erik Tollerud"
         ],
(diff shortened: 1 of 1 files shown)


eteq (approved):
Yep, agreed this is no longer necessary! LGTM

pllim:
👍 And there is no saying that in the future bot won't be on CoCo... 😏 🤖

adrn:
@pllim haha! I didn't even consider that 😅

PR #487, opened 2022-04-21
Updating roles to add APE0 links when relevant

eteq:
While looking at #486 I realized we should probably explicitly note which roles are detailed in APE0.  So this PR does exactly that, adding a couple of links back to APE0 for the Coco and the ombudsperson (the voting members are already linked to APE0.)


pllim (approved):
Very APE-propriate.

adrn (approved):

PR #492, opened 2022-05-17
Add PyAutoLens as AstroPy Affiliated Package

Jammy2211:
We are requesting affiliated status for our package PyAutoLens :).

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -1,6 +1,24 @@
 {
     "packages": [
         {
+          "name": "PyAutoGalaxy",
+          "maintainer": "James Nightingale <james.w.nightingale@durham.ac.uk>",
+          "stable": true,
+          "home_url": "https://pyautogalaxy.readthedocs.io/en/latest/",
+          "repo_url": "https://github.com/Jammy2211/PyAutoGalaxy",
+          "pypi_name": "https://pypi.org/project/autogalaxy/",
+          "description": "Analysing the morphologies and structures of large multi-wavelength galaxy samples. ",
+          "review": {
+             "functionality": "To be filled out by the reviewer",
+             "ecointegration": "To be filled out by the reviewer",
+             "documentation": "To be filled out by the reviewer",
+             "testing": "To be filled out by the reviewer",
+             "devstatus": "To be filled out by the reviewer",
+             "python3": "To be filled out by the reviewer",
+             "last-updated": "To be filled out by the reviewer"
+          }
+        },
+        {
           "name": "BayesicFitting",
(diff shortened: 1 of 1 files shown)


affiliated/registry.json
```diff
@@ -1,5 +1,23 @@
 {
     "packages": [
+        {
+          "name": "PyAutoLens",
+          "maintainer": "James Nightingale <james.w.nightingale@durham.ac.uk>",
+          "stable": true,
+          "home_url": "https://pyautolens.readthedocs.io/en/latest/",
+          "repo_url": "https://github.com/Jammy2211/PyAutoLens",
+          "pypi_name": "https://pypi.org/project/autolens/",
+          "description": "Analysis, modeling and simulations of strong gravitational lenses.",
```
dhomeier, line 10:
```suggestion
          "pypi_name": "autolens",
          "description": "Analysis, modeling and simulations of strong gravitational lenses.",
          "coordinated": false,
```
Only use "project" name on PyPI; + required field.

dhomeier (commented):
Thank you for proposing this package as an affiliated package to the Astropy project. I am presently searching for a reviewer to provide a detailed review; in the meantime I am only asking to make the suggested formal/formatting adjustments to the registry.

dhomeier:
Apologies for the delay as it took some time to find a referee familiar with this field of research. I am however happy to confirm that your package is now under review and we'll post the results of the review here and on the mailing list.
In the meantime I would merely suggest that you add a CITATION file to the main repo containing the principal information of the Citations section in your docs on how to cite your package, so it would be more visible to users of the package.

Jammy2211:
I think with PyAutoGalaxy, the process began on there and never got round to PyAutoLens (which is a similar review as they are child / parent packages).

It sounds like you're getting the proess on PyAutoGalaxy going again, so lets get that rolling and then we can sort PyAutoLEns -- you can either try do them both simultaneouisly or one after another.

Part of its my fault -- I got buried in doing actual science for a few years and had to take a back seat on development :)

hamogu:
I wouldn't call "doing science" a fault. Isn't that why we write all this software in the first place?

hamogu:
Unless they are packaged together, we should review them as separate packages I think; they have different repros, different DOIs, etc.

hamogu:
I'm closing this based on this discussion and expect to see it in PyOpenSci some time after PyAutoGalaxy! 
Don't be surprised that there are still other similar PRs open here. Our goal is transfer them to PyOpenSci one-by-one and not all at once because that would make it harder to find reviewers. PyAutoGalaxy looks the most ready, so I'm starting with your two packages!

PR #496, opened 2022-08-01
Add new voting members elected in July 2022 election

adrn:
Adding Eero and Ricky, as elected in the July election.

### DIFF
--- a/team.html
+++ b/team.html
@@ -158,6 +158,7 @@
       <li>Stuart Littlefair</li>
       <li>Brett Morris</li>
       <li>Stuart Mumford</li>
+      <li>Ricky O'Steen</li>
       <li>Aarya Patil</li>
       <li>Adrian Price-Whelan</li>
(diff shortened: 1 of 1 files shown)


team.html
```diff
@@ -158,6 +158,7 @@ <h3 id="votingmembers-active">Active Voting Members<a class="paralink" href="#vo
       <li>Stuart Littlefair</li>
       <li>Brett Morris</li>
       <li>Stuart Mumford</li>
+      <li>Ricky O'Steen</li>
```
pllim, line 161: @rosteen , do you want your first name here like this or the other way?

pllim (approved):

team.html
rosteen: This is fine, thanks for checking.

mwcraig:
Thanks for the review and congrats and thanks again for running!

PR #502, opened 2022-09-12
Remove APW from two roles

adrn:
These are roles that I either will not have time to do in the next year, or haven't been doing effectively.


adrn:
As a maintainer, and the person in question, I'm going to self approve and merge!

pllim:
😿

PR #504, opened 2022-11-04
Add asdf-astropy to main team page

WilliamJamieson:
I noticed that `asdf-astropy` is not listed on the https://www.astropy.org/team.html page under the `Coordinated package maintainer` heading, but it is listed under `Coordinated Packages` on https://www.astropy.org/affiliated/index.html. This PR resolves this difference.

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -654,6 +654,14 @@
                     "Larry Bradley",
                     "Adam Ginsburg"
                 ]
+            },
+            {
+                "role": "asdf-astropy",
+                "people": [
+                    "Nadia Dencheva",
+                    "Perry Greenfield",
+                    "William Jamieson"
+                ]
             }
         ],
         "responsibilities": {


pllim (approved):
Good catch! But I don't know how to preview this. Make sure it is under "Coordinated package maintainer" section.

pllim:
cc @astropy/coordinators

pllim:
That error been there forever. I just ignore it.

WilliamJamieson:
> That error been there forever. I just ignore it.

Should we let them know about the issue? It appears that they don't support https.

pllim:
@WilliamJamieson , sure. I just never cared enough to follow up and seems like the website maintainer never did either.

hamogu (approved):

hamogu:
The Coco had previously voted to make asdf coordinated, so the fact that it's not listed here is clearly just an othersight. Thanks for catching!

Issue #167, opened 2017-07-05
Role archives and history page


kelle:
From @astrofrog: 

> Somewhere (doesn't have to be too obvious necessarily) we should keep a list of people who were previously in roles, so that there is an official place we record this (because I think we should recognize past as well as current contributions)

apooravc:
@eteq @kelle where can I get the list of people previously in roles?

astrofrog:
@apooravc - there is no list currently, but the only ones I can think of are:

- Perry Greenfield (coordination committee)
- Mike Droettboom (lead for io.votable, wcs)
- Erik Bray (lead for io.fits)

pllim:
There is a history page now but not sure about "role archive".

PR #518, opened 2023-03-06
Roles: GSoC under OpenAstronomy now

pllim:
Astropy has not been active for several years now. And even if it becomes active again, it would be under OpenAstronomy. For instance, see https://openastronomy.org/gsoc/gsoc2023/#/mentors for GSoC 2023.

Or do you think we should keep the role but link directly to OpenAstronomy page? I don't see a "roles" page that that is not by year. Maybe I missed it?

Affected people in the removed listings:

* @bsipocz 
* @eteq 
* @mirca

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -109,20 +109,6 @@
         }
     },
     {
-        "role": "Astropy GSoC coordinator",
-        "url": "Astropy_GSoC_coordinator",
-        "people": [
-            "Brigitta Sip\u0151cz",
-            "Erik Tollerud",
-            "Z\u00e9 Vin\u00edcius"
-        ],
-        "role-head": "Astropy GSoC coordinator",
-        "responsibilities": {
-            "description": "Coordinate the participation of Astropy in the Google Summer of Code Project",
-            "details": []
-        }
-    },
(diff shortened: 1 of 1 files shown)


pllim:
Also related: https://github.com/astropy/astropy.github.com/issues/167 -- Do we move this to some archive instead? Where?

eteq:
I think of this role as "Astropy liason to the GSoC part of OA".  So I don't think we should remove it, although it isn't relevant in years like this one, where we have no GSoC mentorship capacity.

eteq:
But agreed since we haven't done it for a few years maybe it should just be declared unfilled and clarify in the role description that   it might be unfilled as long as we aren't active in GSoC?

eteq:
After discussion at the coordination meeting, the consensus is that, even though this role has sometimes been needed in the project, it's not in operation right now. We might bring it back in the future but right now it doesn't seem like it makes sense to keep this between the ambsence and existence of OA as a starting point.

eteq:
(discussion included @pllim @hamogu @weaverba137 @aaryapatil @saimn)

PR #520, opened 2023-03-06
Roles: Merge documentation infrastructure and testing infrastructure

pllim:
Documentation and testing infrastructure are closely related and both are actually managed by the same set of people in reality. say, it is tough to maintain `pytest-astropy` without also touching `sphinx-astropy` since doctest tie them together. And both documentation and tests are done in the same manner (e.g., the CI checks for PR).

If simply merging them is too controversial, we can also try to have each of them as "sub-role" but shared by the same set of people. 🤷 

Affected people in the merged listings:

* @saimn 
* @pllim 
* @astrofrog 
* @bsipocz 

TODO:

- [x] Link to chart from https://github.com/astropy/astropy-project/issues/118 once it is updated and in the repo.
- [ ] Add sub-roles for the many things that we actually work on

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -239,6 +239,7 @@
                 "Overseeing content (although primary responsibility for content lies with subpackage maintainers)",
                 "Managing issues/pull request for the Astropy core package regarding testing infrastructure",
                 "Managing issues/pull requests in the repositories containing the testing plugins, and determining when new plugins are required",
+                "Set up and maintain continuous integration services",
                 "Maintaining the 'metapackage' with the testing machinery (pytest-astropy at the time of this writing)",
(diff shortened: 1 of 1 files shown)


pllim:
Given @hamogu's mention of "phone book," I could also expand this PR to add all the sub-roles for infrastructure. I mean, right now they are all the same people, but I guess it is not guaranteed to remain so when/if we finally have more staff to help out.

Please see https://github.com/astropy/astropy-project/issues/118 .

astropy_infrastructure_20230307

hamogu:
This will be rebuild as a separate PR, where infrastrucure roles will be specified in more detail

PR #521, opened 2023-03-06
Roles: Add specreduce to Coordinated section

pllim:
`specreduce` is listed as Coordinated under https://www.astropy.org/affiliated/index.html#coordinated-package-list but not listed in `roles.json`. Here, I simply copy over the "Maintainer(s)" listed over there to this JSON file. Whether they are really actively maintaining or not, I do not know. If this listing is no longer accurate, please tell me the current maintainers for `specreduce`. Thanks!

Affected people in the added listing:

* @tepickering
* @eteq

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -662,6 +662,13 @@
                     "Perry Greenfield",
                     "William Jamieson"
                 ]
+            },
+            {
+                "role": "specreduce",
+                "people": [
+                    "Tim Pickering",
+                    "Erik Tollerud"
+                ]
             }
         ],
         "responsibilities": {


eteq (approved):

roles.json
```diff
@@ -662,6 +662,13 @@
                     "Perry Greenfield",
                     "William Jamieson"
                 ]
+            },
+            {
+                "role": "specreduce",
+                "people": [
+                    "Tim Pickering",
```
eteq, line 669:
```suggestion
```

hamogu:
I think we can do that "add people" in the same PR and follow the regular process for this PR.

roles.json
hamogu: Tim agreed verbally

pllim: Just for completeness, Tim is already on https://github.com/orgs/astropy/teams/specreduce-maintainers , so I don't see the problem. 

If Kyle should be promoted to this role, please open a separate PR and we can go through the usual procedures for adding people. Thanks!

PR #523, opened 2023-03-06
Roles: Remove package template

pllim:
This one might want to wait for when we officially retire the template in favor of OpenAstronomy, or should we go ahead?

* https://github.com/astropy/package-template/issues/519

Affected people in the removed listing:

* @larrybradley
* @Cadair 
* @astrofrog 
* @bsipocz

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -283,26 +283,6 @@
         }
     },
     {
-        "role": "Package template maintainer",
-        "url": "Packagetemplate_maintainer",
-        "people": [
-            "Larry Bradley",
-            "Stuart Mumford",
-            "Thomas Robitaille",
-            "Brigitta Sip\u0151cz"
-        ],
-        "role-head": "Package-template maintainer",
-        "responsibilities": {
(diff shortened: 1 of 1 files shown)


eteq:
I think we absolutely need a role in the project that maintains this.  I think we may want to re-imagine this as more like an "OA liason" though?  But I think it's important we recognize that this is a real part of the project.

pllim:
I am still of the opinion that this role should be removed until we have a concrete plan w.r.t. template coordination with OpenAstronomy and people who actually have time to do the work. I already archived package-template, so the existing description no longer applies.

mhvk:
I agree with @pllim that the role as described is gone, so it should be removed. I do also agree with @eteq that we should have a role like this (even if just for a transition document), but given that we have neither a role description nor a volunteer, I think that should be a separate issue -- no reason not to remove a defunct role just because we need a similar one (unless of course someone adds a description and volunteers for the role in the next day or two).

hamogu:
I agree with @mhvk : We should remove this role for now. If someone (e.g. @eteq ) then they can suggest a new role (ideally with a person to fill it) they can open a new PR.

PR #530, opened 2023-04-28
Roles: Update release team members

pllim:
In the past several releases (and we only do a release every 6 months, so it's been a while), only @astrofrog and @saimn are alternating, so based on that, I am updating the Release Team membership.

Affected people in this listing:

* @embray 
* @bsipocz 

And, I have to confess that I accidentally already updated the actual "Astropy Project Release Team" GitHub membership. I thought this was already done over here but I must have imagined it. So, if this PR ends up changing, I'll update on GitHub side accordingly. I take full responsibility for this mistake and I apologize for any inconvenience caused.

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -366,10 +366,8 @@
         "role": "Release team",
         "url": "release_team",
         "people": [
-            "E. Madison Bray",
             "Simon Conseil",
-            "Thomas Robitaille",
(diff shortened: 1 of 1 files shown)


pllim:
This has been open for a while and it is still true that Tom R and Simon C are the only ones actively doing releases for `astropy` now, so I am merging this.

Issue #519, opened 2023-03-06
Roles: Audit all existing titles and membership


pllim:
Motivation: Over the years, the Project has evolved more quickly than `roles.json` (that renders to https://www.astropy.org/team) can be updated. As a result, there are now roles that do not exist anymore (or inactive) and people who do not really fill existing roles anymore. I would like to submit a series of PRs to update this listing, or at least put the roles in discussion.

Depends on:

* Developer survey results (2023 and maybe also 2022) -- who replied and who MIA?
* Community discussions (on this issue, on related PRs, mailing lists, Slack, etc)

Also might be relevant:

* #167

Existing roles (as of 2023-03-06):

- [x] Coordination committee member -- This is decided by election. No further action needed.
- [x] Ombudsperson -- Still true. No further action needed.
- [ ] Community engagement coordinator -- Do we really need so many sub-roles? Can this be all merged into one when we hire a proper community manager?
- [ ] Astropy GSoC coordinator -- #518 
- [ ] Learn Team -- Do we really need so many sub-roles? Are these even still active? Can this be simplified?
- [x] Finance committee member -- This committee is active. No further action needed.
- [ ] Documentation infrastructure maintainer -- Can this be merged with "Testing infrastructure maintainer"? I am active in both and they do overlap a lot.
- [ ] Affiliated package review editor -- They are active. No further action needed.
- [ ] Astropy.org web page maintainer -- I don't know what this means. If this is simply who holds the DNS account, I think this is overkill and can be removed. If not, please clarify what this role actually does.
- [ ] Astropy-helpers maintainer -- #517 
- [ ] DevOps and Operations Specialist -- We tried to hire an official person to do this but that person is no longer active and the position has lapsed. I say we remove this because this is now split into other roles already mentioned (e.g., infrastrucure, core).
- [ ] Testing infrastructure maintainer -- Can this be merged with "Documentation infrastructure maintainer"? I am active in both and they do overlap a lot.
- [ ] Package template maintainer -- We plan to retire the template in favor of OpenAstronomy's, so we should remove this role because it is no longer actively maintained. See https://github.com/astropy/package-template/issues/519
- [ ] CI-helpers maintainer -- #517 
- [ ] Distribution coordinator -- It is very hard to vet this one since distro activities happen outside the purview of Astropy. I know some of these people are active while hardly heard from the other ones, though that does not mean they are not doing anything either. Also, other distros out there might be distributing `astropy` also but we don't list them. Ideas welcome.
[...]

PR #535, opened 2023-05-09
Roles: Move Finance Committee to after Ombudsperson

pllim:
As discussed and agreed upon at Astropy Coordination Meeting 2023 (myself, @eteq, @hamogu, @weaverba137, @aaryapatil, @saimn), Finance Committee shall be moved up to after Ombudsperson. This is because given that the committee now oversees multiple grants for the Project and so on, that would be a logical (and hopefully uncontroversial) place to put it.

Affected people (the Finance Committee):

* @kelle 
* @aaryapatil
* @jdswinbank 
* @eteq 

Out of scope: Re-ordering of all other roles.

This is part of #519

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -56,6 +56,26 @@
         }
     },
     {
+        "role": "Finance committee member",
+        "url": "finance_committee_member",
+        "people": [
+            "Kelle Cruz",
+            "Aarya Patil",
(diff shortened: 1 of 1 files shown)


eteq:
Just noticed this was not merged, merging now!

PR #538, opened 2023-05-17
Removing myself from Facebook community role

eteq:
This PR is me removing myself from this role - in practice I do not use Facebook anymore so I cannot be effective in this role (although I still occasionally can pop in to leave comments when pinged).

cc @pllim who reminded me I can do this myself after I said it in the roles survey :wink:

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -75,8 +75,7 @@
             {
                 "role": "Facebook",
                 "people": [
-                    "Kelle Cruz",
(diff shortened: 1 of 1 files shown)


eteq:
Normally I try to strictly follow the no-merging-own-PR rule, but I guess it's ok to relax that in this case given it's me removing myself from a role and has no technical consequences

PR #536, opened 2023-05-09
Roles: Remove Madison due to inactivity

pllim:
The SOSS contract has ended. Furthermore, Madison has not been active in the assigned roles for a while. Therefore at Astropy Coordination Meeting 2023 (myself, @eteq, @hamogu, @weaverba137, @aaryapatil, @saimn), we decided to remove them from those roles.

Affected person:

* @embray 

Keep in mind that Madison still had made very valuable contributions and those are reflected in https://www.astropy.org/credits.html . They are also still a Voting Member.

Part of #519 , half the changes here is also subset of #530

Issue #412, opened 2020-10-26
Reorder roles


astrofrog:
At the moment the roles are in a bit of a random order, as several infrastructure ones are separated. Things like devops, docs infra and testing infra should probably be listed together and before some more specific things like astropy-helpers and ci-helpers.

pllim:
Took a stab at https://github.com/astropy/astropy.github.com/pull/609 . Please discuss there. Thanks!

PR #539, opened 2023-05-22
Roles: Add sub-roles for infrastructure and other related updates, remove Madison due to inactivity

pllim:
Another stab at the infrastructure stuff in #519 .

* There is now a proper Infrastructure Team with sub-roles.
  * The existing ones are pretty comprehensive to begin with, so I mainly just move titles around.
  * Folded the website title into infrastructure, because I think it is.
* For some roles, I clarified and expanded their responsibilities. Most are based on what I already do, but I wish people with those roles would do instead.
* Added a new unfilled role for "metrics". This is because I think benchmarks and dashboard are important but they are broken, and we need a dedicated role for them going forward, if we ever want working benchmarks and dashboard again.
* Merged "Core astropy package maintainer (general)" and "Core astropy package maintainer (sub-package)", both now under "Core astropy package maintainer", but they retained their sub-roles. I needed to do this here because I needed to clarify some overlap between the "general" core maintainer with the "DevOps Specialist".

Incorporated later:

* #536 

Preview: https://output.circle-artifacts.com/output/job/1197d6ec-ae8c-4a71-a3d2-8da7a525b1dc/artifacts/0/html/team.html

Out of scope: Removing/added names besides the one. We have agreed at the Astropy Coordination Meeting 2023 that names will be touched at a later stage.

Out of scope: Reordering the roles. That will also happen later as part of #412.

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -207,29 +207,108 @@
         }
     },
     {
-        "role": "Documentation infrastructure maintainer",
-        "url": "Documentation_infrastructure_maintainer",
+    "role": "Infrastructure Team",
+    "url": "Infrastructure_team",
+    "role-head": "Infrastructure Team",
+    "sub-roles": [
+        {
+        "role": "DevOps and Operations Specialist",
+        "people": [
+            "Pey Lian Lim",
+            "Brigitta Sip\u0151cz"
+        ]
+        },
+        {
+        "role": "CI Infrastructure",
         "people": [
             "Simon Conseil",
             "Pey Lian Lim",
             "Thomas Robitaille",
             "Brigitta Sip\u0151cz"
-        ],
-        "role-head": "Documentation infrastructure maintainer",
-        "responsibilities": {
-            "description": "Maintain the <a href='http://docs.astropy.org/en/stable/index.html'>Astropy documentation</a> website, including:",
-            "details": [
-                "Managing the Sphinx infrastructure",
-                "Implementing changes and improvements to the documentation website",
-                "Overseeing content (although primary responsibility for content lies with subpackage maintainers)"
-            ]
+        ]
+        },
+        {
+        "role": "Documentation Infrastructure",
+        "people": [
+            "Simon Conseil",
+            "Pey Lian Lim",
+            "Thomas Robitaille",
+            "Brigitta Sip\u0151cz"
+        ]
+        },
+        {
+        "role": "Metrics",
+        "people": ["Unfilled"]
+        },
+        {
+        "role": "Website and Data",
+        "people": [
+            "Adrian Price-Whelan",
+            "Erik Tollerud"
+        ]
         }
+    ],
+    "responsibilities": [
+        {
+        "subrole-head": "DevOps and Operations Specialist",
+        "description": "Ensure the smooth running of the project:",
+        "details": [
+            "Perform initial triaging of issues and pull requests, including moving between repositories in the project.",
+            "Perform first responder duties across the project, including guiding first-time contributors.",
+            "Monitor the health of the project in terms of development and operations; if necessary, also alert the relevant project members of any issues and/or formulate responses.",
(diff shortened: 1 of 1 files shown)


pllim:
Hmm what happened to Giles preview?

roles.json
```diff
@@ -207,29 +207,110 @@
...
-        "role": "Documentation infrastructure maintainer",
-        "url": "Documentation_infrastructure_maintainer",
+    "role": "Infrastructure Team",
+    "url": "Infrastructure_team",
+    "role-head": "Infrastructure Team",
+    "sub-roles": [
+        {
+        "role": "DevOps and Operations Specialist",
+        "people": [
+            "E. Madison Bray",
```
hamogu, line 217: Is Madison still doing it? I know we said at the Coordination meeting that we'll separate the steps of "defining roles better" and "adding/removing people" separately, so not a requirement for this PR, but in clear cases sould also remove names right now.

roles.json
```diff
@@ -207,29 +207,110 @@
...
+        ]
+        },
+        {
+        "subrole-head": "CI Infrastructure",
+        "description": "Lead development and maintenance of the continuous integration (CI) testing infrastructure for Astropy, including:",
+        "details": [
+            "Set up and maintain continuous integration services.",
+            "Managing issues/pull request for the astropy core package regarding testing infrastructure.",
+            "Managing issues/pull requests in the repositories containing the testing plugins, and determining when new plugins are required.",
+            "Maintaining the 'metapackage' with the testing machinery (pytest-astropy at the time of this writing).",
```
hamogu, line 272:
```suggestion
            "Maintaining the 'metapackage' with the testing machinery (e.g. pytest-astropy).",
```
While, strictly speaking there is a git commit that specifies "at the time of this writing", on the website you don't see that. So, I suggest are more generic wording even at the risk of that being imprecise at some point in the future.

roles.json
```diff
@@ -207,29 +207,110 @@
...
+        ]
+        },
+        {
+        "subrole-head": "Metrics",
+        "description": "Manage the tools needed for project metrics:",
+        "details": [
+            "Run benchmarks (mostly the core package) and publish results at a frequent cadence.",
+            "Maintain the benchmark tools.",
+            "Managing pull requests to the benchmark repository.",
+            "Stretch goal: Lead the effort to revive project dashboard."
```
hamogu, line 295: I would either decide that it's part of this role or not instead of listing "stretch goals". If it's not done (and the role is unfilled anyway right now) maybe err on the side of listing only stuff we really want to get done.

roles.json
```diff
@@ -207,29 +207,110 @@
...
+        ]
+        },
+        {
+        "subrole-head": "Website and Data",
+        "description": "Manage the <a href='https://astropy.org'>astropy.org</a> website, including:",
+        "details": [
+            "Maintaining contributor/roles list.",
```
hamogu, line 302: I suggest to take this out. I see that really as a Coco responsibility since it's the Coco that puts people in roles and that decides when people are not in roles any longer. So, it should not be on the website maintainers to maintain that - the Coco has to communicate and keep track.

roles.json
pllim: Okay, I incorporated https://github.com/astropy/astropy.github.com/pull/536 here.

roles.json
pllim: Accepted your change. Thanks!

roles.json
pllim: I really think the project dashboard needs to be revived at some point. So I left this in but removed "stretch goal".

roles.json
pllim: Removed the first line.

pllim:
@hamogu , thanks for the comments. I believe I have addressed them. I also resolved the merge conflict and the preview magically works again!

pllim:
Things have changed a little and there is conflict, so I am closing this without merge and will revisit with a new (and hopefully less controversial) PR.

PR #552, opened 2023-10-19
Updated installation instructions

astrofrog:
This updates the front page installation instructions:

* No longer recommend the Anaconda Python Distribution as being the best way - as in our main install instructions, pip should be first.
* Include ``--upgrade`` flag for pip because might as well to ensure the latest version (mentioned in https://github.com/astropy/astropy/issues/15487)
* Include ``[recommended]`` for pip since, well, it is recommended!
* Use conda-forge for conda due to https://github.com/astropy/astropy/issues/15122 - even though right now defaults is up to date again, it might not be in future. Conda-forge will always be more likely to be up-to-date.

I think it's important to fix this as these are our most visible user-facing installation instructions.

### DIFF
--- a/index.html
+++ b/index.html
@@ -118,19 +118,12 @@ window.onload = function() {
 
 	<section class="install">
 		<h1 id="install-astropy">Install Astropy<a class="paralink" href="#install-astropy" title="Permalink to this headline">¶</a></h1>
-	  The <a href="https://www.anaconda.com/download/">Anaconda
-	    Python Distribution</a> includes astropy and is the recommended way to install both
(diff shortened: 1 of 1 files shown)


hamogu (approved):

mhvk:
Thanks, much better!

dhomeier:
Remided me that [this](https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community) just came up on pyOpenSci, but I did not find the link in time. I don't know if this would have any effect on the default channels or just concerns the solver engine under the hood; in that case it would only affect recommendations regarding conda vs. miniconda vs. mamba...

PR #556, opened 2023-10-22
Remove affiliated package decisions from Coco

hamogu:
The Coco has delegated this responsibility to dedicated affiliated package editors years ago, thus I suggest to not list is as Coco role at this stage.
In fact, we're about to pass that over to PyOpenSci and thus might not need the affiliated package editor role either in the near future, but that's for a future PR.


pllim (approved):
Good catch. Having been on both sides, I totally agree that CoCo has not been doing this, so I am just going to merge.

PR #558, opened 2023-10-28
Bugfix: bibtex for astropy 2022

bmorris3:
I added the bibtex entries for astropy to a manuscript today, and checked the compiled references, which looked like this:

Screenshot 2023-10-28 at 14 15 30

Note the unrecognized `apj` in the top astropy reference. This is because we need another escape character in the string on the webpage. With the extra escape, you'd get what we intend:

Screenshot 2023-10-28 at 14 15 48

This PR simply adds the extra `\`.


pllim (approved):
Thanks!

PR #562, opened 2023-11-27
Update team page

saimn:


team.html
```diff
@@ -465,7 +480,7 @@ <h3 id="core-package-contributors">Core Package Contributors<a class="paralink"
 			<li>Luigi Paioro</li>
 			<li>Luke G. Bouma</li>
 			<li>Luke Kelley</li>
-			<li>Luz Paz</li>
+			<li>luz paz</li>
```
pllim, line 483: Why the de-capitalization here?

saimn: That's the script output, it may have been fixed manually.

dhomeier:
Wondering if we should somehow format the contributors list such as to indicate whether something is a real name or a github handle.

pllim:
@dhomeier , that would not be easy also I don't think that is necessary. We should represent them as they want to be represented in their GitHub profile.

PR #576, opened 2024-02-19
William Jamieson is stepping down from several roles

WilliamJamieson:
I am stepping down from several of my rolls with the Astropy project. This is because I am moving away from working at STScI to working for a private company. Due to this change my time and ability to contribute to astropy will be much more limited. For now I believe that I can continue to act as an `astropy.modeling` maintainer as a volunteer in my personal time.

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -217,7 +217,7 @@
         },
         {
             "name": "asdf-astropy",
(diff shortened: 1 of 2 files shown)


pllim (approved):
Given that asdf-astropy is Coordinated, not core, I don't think it needs a two-week period. Looks like Coordinated maintainers are listed in two different places in Brett Graham is already in one of them anyway.

Sorry to see you go!

PR #577, opened 2024-03-12
DOC: updating pyvo maintainer list

bsipocz:
To reflect reality. Accompanying task is to update the github team and revise the accesses to the repo (as it's currently a mixed of individuals rather than team-based)

cc @pllim

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -356,7 +356,7 @@
         },
         {
             "name": "PyVO",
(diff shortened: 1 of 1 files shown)


pllim:
Thanks, @bsipocz ! Is it possible to ping all the affected people (removed and added), so they are aware this is happening?

pllim:
I think I cleaned up `pyvo` access by teams. `pyvo` should now have 2 different teams, one with admin access and another with write access. The new people on this list would have admin access. However, I cannot merge this until I know Markus' GitHub username. Thanks!

PR #581, opened 2024-04-05
Update the Help page in response to Astropy Community Engagement survey

taldcroft:
This PR is intended to take on the results of the [Astropy Community Engagement survey](https://astropy-report.orgmycology.com/sections/exec_summary). In particular this PR is in response to comments related to [Communication](https://astropy-report.orgmycology.com/sections/exec_summary#communication-channels) and [astropy communication details](https://github.com/orgmycology/astropy_report/blob/main/sections/3_communication.md).

In any large community there will be differences of opinion, but the highlights this PR tries to address are below.

#### Response: remove Twitter/X and Facebook from the recommended forums. 

The community has said "less is more". Instead of providing as many options as possible, let's focus on the best options.

Twitter was never a good match for answering technical questions. The move to X has made this platform less attractive as an option.

The Facebook channel has seen a significant drop in engagement (4 posts in the last month) and only a small fraction of developers have Facebook accounts (1 of 9 at the last dev telecon).

#### Response: highlight Slack as a user forum with a new `#community-help` channel

Slack is highlighted in many places in the report, with a generally positive slant. We recognize the limitations of this platform (most notably the lack of open search) but we should likewise recognize the success of Slack for the developer community. As a platform for having technical discussions involving code, often in real time, it works well.

There has been resistance to opening up Slack for community user questions since that might result in "helpdesk" questions being posted to developer channels. The idea here is to make a new channel `#community-help` which is intended as the *one place* in the astropy Slack for user questions. The `community` prefix specifically highlights that this is a community channel and may not have developer involvement.

It is worth noting that the `#table` channel is already dominated by help questions, including questions from developers that consistently argue that Discourse is where such questions should be posted. In this one channel there has been no problem with the same question being repeated due to the lack of open search capabilities.

### Other changes

This PR generally removes any statements about forum activity and expected response time. Those metrics evolve with time and with a reduced set of options, there is no need to provide this information to users.

Remove plans to redirect questions to Discourse. I'm not aware if this has happened, but our main help page should only discuss implemented process, not plans. :smile:


pllim:
> One unrelated commit fixes a javascript error highlighted by my IDE. I'm not sure about the fix so someone with knowledge should review that (or we can just take it out).

Maybe should be a separate PR, since if that fix is real, should be merged sooner than later. Thanks!

help.html
```diff
@@ -73,54 +73,32 @@
...
-		contributors use Slack for everyday discussions. Feel free to use this
-		forum if you have any questions about contributing to Astropy. The best place to ask questions about how to use Astropy is the
-		<a href="https://community.openastronomy.org/c/astropy/8" title="">Open Astronomy Discourse</a> server.
-		Please browse all the available channels before posting and try to choose the most
-		relevant one.
-		</li>
+		<li> <a href="https://astropy.slack.com">Astropy Slack workspace</a> (<a
+            href="http://joinslack.astropy.org">get an account</a>) - This forum is primarily
+        focused on developing and maintaining the Astropy project, but the
+        <code>#community-help</code> channel is a good place for users to seek help from the
```
pllim, line 87: Thanks for the updates!

I am still worried that users would think any answers given are canon and vetted by the devs, though that is actually quite impossible to promise. We would at least need a moderator for that channel, preferably the Community Manager. Most devs who want to minimize noise might not even join this channel. And some people just opposed to Slack in general. Might need more input from the community.

I cross-linked this PR to:

* https://github.com/astropy/astropy-project/issues/349

taldcroft: > I am still worried that users would think any answers given are canon and vetted by the devs, though that is actually quite impossible to promise. We would at least need a moderator for that channel...

I do not see how Slack is any different from Discourse, FB, Twitter, or email in this regard. To my knowledge none of these forums has an active moderator that ensures technical accuracy of information.

In general the Astropy project provides no guarantees of accuracy or correctness in any of our communications with users. This is a community project and we do not distinguish between devs (i.e. people with commit rights) and community members when it comes to responding to user questions.

> Most devs who want to minimize noise might not even join this channel.

That's fine! The channel is explicitly named `community-help` to highlight that it is a community-driven forum and developers need not participate, just as there is no requirement that they participate in Discourse, FB, X, or emails.

> And some people just opposed to Slack in general.

No worries, there are other options available to people opposed to Slack. However, there are also people (as evidenced in the survey) are comfortable with Slack and find this a good option.

dhomeier:
Thanks for tidying this up! Not just clarifying the best recommendations in response to the survey and dev telecon, but also making the whole page a lot more concise.

> Remove plans to redirect questions to Discourse. I'm not aware if this has happened, but our main help page should only discuss implemented process, not plans. 😄

I was wondering if officially endorsing Discourse over the other channels was not already part of that redirection, but no need to elaborate on that on the help page itself, and anyway with the new setup it no longer really applies, as Slack and Discourse will be presented with more or less equal priority (at least that is my understanding now).

The only thing I am unsure about is whether erasing all mentions of FB and Twitter/X is the best way to go; while I'd gladly let them sink into oblivion, maybe we should just note that we are aware of them and would not recommend turning to one of these old channels – otherwise someone may come in any day to notify us of this helpful forum they have just discovered. 😉

taldcroft:
> The only thing I am unsure about is whether erasing all mentions of FB and Twitter/X is the best way to go; while I'd gladly let them sink into oblivion, maybe we should just note that we are aware of them and would not recommend turning to one of these old channels – otherwise someone may come in any day to notify us of this helpful forum they have just discovered. 😉

If that ever happens that could be an opportunity to reevaluate the ecosystem of community help by discussing the user's experience. Maybe FB has become a useful and active forum?

In the meantime my I think this page should include only our *recommended* forums. Less is more. :smile:

pllim:
Since the suggestion seems very trivial and is more correct, I took the liberty to accept it. Thanks!

PR #583, opened 2024-04-06
Fix unterminated string literal

taldcroft:
I'm not entirely sure this is correct but it looks reasonable to me.


pllim (approved):
Looks about right to me too. Thanks!

PR #584, opened 2024-04-29
Add strategic planning role

eteq:
This came out of some discussions the CoCo has been having the last few months on whose role it is to do big picture planning.  We eventually settled on the reality of the current situation: that it has historically been something the CoCo does, but it is not *necessary* that this be so, and we should make it an explicit role so others can participate if they want.

As a starting point I named the people as just "the CoCo members" since that's the current reality, but that could change if others want to lean in on this topic.

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -55,6 +55,25 @@
         }
     },
     {
+        "role": "Strategic Planner",
+        "url": "strategic_planner",
+        "people": [
+            "All Coordination Committee Members"
+        ],
+        "role-head": "Strategic planner",
+        "responsibilities": {
+            "description": "Do long-term planning for the Astropy Project as a whole, including:",
+            "details": [
+                "Shape and provide a vision for next major steps in the Project, in consultation with the whole Astropy community.",
+                "Maintain a large-scale view of the project (both from developer and user perspectives).",
+                "Provide points of contact for external stakeholders on the above (with the CoCo).",
+                "Monitor efforts on the topic above to ensure they are being done as imagined.",
+                "Write grant proposals (with the Finance Committee).",
+                "Oversee the Astropy Project Roadmap."
+            ]
+        }
+    },
+    {
         "role": "Finance committee member",
         "url": "finance_committee_member",
         "people": [
--- a/roles.json
+++ b/roles.json
@@ -55,7 +55,7 @@
         }
     },
     {
-        "role": "Strategic Planner",
+        "role": "Strategic planning committee member",
         "url": "strategic_planner",
         "people": [
             "All Coordination Committee Members"
@@ -65,11 +65,12 @@
             "description": "Do long-term planning for the Astropy Project as a whole, including:",
             "details": [
                 "Shape and provide a vision for next major steps in the Project, in consultation with the whole Astropy community.",
-                "Maintain a large-scale view of the project (both from developer and user perspectives).",
+                "Maintain an overview of all the project activities (both from developer and user perspectives).",
                 "Provide points of contact for external stakeholders on the above (with the CoCo).",
-                "Monitor efforts on the topic above to ensure they are being done as imagined.",
-                "Write grant proposals (with the Finance Committee).",
-                "Oversee the Astropy Project Roadmap."
+                "Monitor project activities to see how well aligned with the roadmap and strategic goals of the project they are, and report back to the CoCo and community.",
+                "Coordinate writing grant proposals (in collaboration with the Finance Committee and wider community).",
+                "Maintain the Astropy Project Roadmap.",
+                "Engage with, and collect input from, the the whole Astropy Community on project vision and priorities."
             ]
         }
     },
--- a/roles.json
+++ b/roles.json
@@ -66,11 +66,11 @@
             "details": [
                 "Shape and provide a vision for next major steps in the Project, in consultation with the whole Astropy community.",
                 "Maintain an overview of all the project activities (both from developer and user perspectives).",
-                "Provide points of contact for external stakeholders on the above (with the CoCo).",
-                "Monitor project activities to see how well aligned with the roadmap and strategic goals of the project they are, and report back to the CoCo and community.",
-                "Coordinate writing grant proposals (in collaboration with the Finance Committee and wider community).",
+                "Provide points of contact for external stakeholders on the above (with the Coordination Committee).",
+                "Monitor project activities to see how well aligned they are with the roadmap and strategic goals of the project, and report back to the community and Coordination Committee.",
+                "Coordinate writing grant proposals (in collaboration with the Finance Committee, Coordination Committee, and community).",
                 "Maintain the Astropy Project Roadmap.",
-                "Engage with, and collect input from, the the whole Astropy Community on project vision and priorities."
+                "Engage with, and collect input from, the whole Astropy Community on project vision and priorities."
             ]
         }
     },


hamogu:
I would suggest to leave this open to the coordination meeting. A lot of strategic planning should happen there! Personally, I'm also interested in these things. However, there is a question of how we organize in practice. In my experience, strategic planning is one of those things that profit from real-time talk (in person or on zoom), unlike, e.g. subpackage maintaining (which can be done fine asynchronosly on GH). 
Maybe dedicate one Coco meeting or one dev telecon every other months (because in practice this overlaps with Coco membership and interested developer membership) to strategic planning and open it to all interested parties?

pllim:
Given the meeting is less than 2 months away, sure, we can wait. I added to ideas here:

* https://github.com/astropy/astropy-project/issues/349#issuecomment-2085419701

pllim (changes requested):
One member expressed desire to wait till Coordination meeting, so I am blocking merge for now. Thanks!

astrofrog:
Just an idea - it seems this would effectively be a steering committee? If so it might be worth considering using a name like this so that it is more familiar for external people, if it becomes an entity that does grow beyond the CoCo.

eteq:
Oh, and @hamogu , re-reading your message, I'm not clear on this: were you saying we should discuss *whether to create this role* at the coordination meeting (as @pllim interpreted it) or that this role should partly be *done* by the coordination meeting (as I interpreted it).  Or both?  Just trying to understand if you were saying action on this should wait on the meeting or not.

(In practice it probably doesn't really matter, since the CoCo is already likely to bring up the strategic items at the meeting as it stands, but I think it will make it slightly easier to recruit folks at the coordination meeting if this roles exists by then since we can say "do you want to join the group?" rather than "do you want to join a group that may or may not exist?")

eteq:
Answering my own question - out-of-band (in the developer telecon) @hamogu said the answer was "both" - which I am totally fine with, just wanted the clarification.

hamogu:
The coordination meeting does fill this role - that's what it is for. So, that should be reflected in the description of the role (e.g. "stratetic planning in between coordination meetings" or "monitoring ... as set by coordination meetings".

But I *also* think that this role should be created at the coordination meeting. It's more open and inclusive to ask for interested parties there, than to pre-fill the role with default names first.

pllim:
So this is more like "strategic plans enforcer" instead of "strategic planning"? 🤔

astrofrog:
Ok so a few thoughts having taken a bit more time to digest this:

* @eteq - I agree that steering committee isn't right because given the range of definitions out there it doesn't really distinguish it from coordination committee.
* Should the role being proposed here be called a 'group' or a 'committee'? For example 'Strategic Planning Committee' or 'Strategic Planning Group'. It's likely it would operate as a group or committee rather than individual people with a role acting independently?
* One of the things that's I've been wondering how we could improve is that indeed beyond the CoCo, any kind of strategic planning is restricted to (at most) once a year for other project members. If strategic planning was done as a committee, it could be a way of increasing participation throughout the year.

@eteq - you said 'could happen at coordination meetings and dev telecons' - I guess we should think about whether to treat the dev telecons as mini coordination meetings in the sense they could also cover strategic stuff and non-development things, or keep the dev telecons be about dev and have separate calls for strategic planning?

roles.json
```diff
@@ -54,6 +54,25 @@
...
+        "role": "Strategic Planner",
+        "url": "strategic_planner",
+        "people": [
+            "All Coordination Committee Members"
+        ],
+        "role-head": "Strategic planner",
+        "responsibilities": {
+            "description": "Do long-term planning for the Astropy Project as a whole, including:",
+            "details": [
+                "Shape and provide a vision for next major steps in the Project, in consultation with the whole Astropy community.",
```
Cadair, line 67: Isn't this just the roadmap?

roles.json
```diff
@@ -54,6 +54,25 @@
...
+                "Maintain a large-scale view of the project (both from developer and user perspectives).",
+                "Provide points of contact for external stakeholders on the above (with the CoCo).",
+                "Monitor efforts on the topic above to ensure they are being done as imagined.",
```
Cadair, line 70:
```suggestion
                "Monitor project activities to see how well aligned with the roadmap and strategic goals of the project, and report back to the CoCo and community.",
```

roles.json
```diff
@@ -54,6 +54,25 @@
...
+        "url": "strategic_planner",
+        "people": [
+            "All Coordination Committee Members"
+        ],
+        "role-head": "Strategic planner",
+        "responsibilities": {
+            "description": "Do long-term planning for the Astropy Project as a whole, including:",
+            "details": [
+                "Shape and provide a vision for next major steps in the Project, in consultation with the whole Astropy community.",
+                "Maintain a large-scale view of the project (both from developer and user perspectives).",
```
Cadair, line 68: This feels vague, could we add something about to what end here?

```suggestion
                "Maintain an overview of all the project activities (both from developer and user perspectives).",
```

Cadair (commented):
Following the discussion at the coordination meeting I think this is a good thing to have, but this committee has a very important remit so I think the wording of this charter is important to get right.

So here's a bunch of word-smithing.

roles.json
```diff
@@ -54,6 +54,25 @@
             ]
         }
     },
+    {
+        "role": "Strategic Planner",
```
Cadair, line 58: I think there is general consensus at the coordination meeting that this *must* be a committee.

```suggestion
        "role": "Strategic Planning Committee",
```

eteq:
A few highlights from the coordination meeting discussion (NOT COMPLETE - these are highlights and distillation from my perspective):

* a lot of agreement strategic planning is important, and that it needs multiple people to do it effectively
* maybe not *too* many, although exactly what that is is a bit ambiguous.
* There is actually a fair amount of interest - in the coordination meeting ~9 hands went up as possibly interested.
* Might need more clarification is that this role is a *devolved* role from the CoCo, since APE0 sort of says this is a CoCo role.
* There is a cross-connection with #89 in that the long-term view of this community is intertwined with how long-term the CoCo terms are.

roles.json
adrn: I missed the roadmap discussion, but it does sound like the roadmap in its current form. But IMO some of the vaguer elements of the roadmap as they currently stand should be removed from the roadmap and added to the charge of this group, and the roadmap should be a set of actionable items with timelines associated.

roles.json
adrn: To be consistent with other names on the team page, I think:
```suggestion
        "role": "Strategic Planning Committee Member",
```

roles.json
```diff
@@ -54,6 +54,25 @@
...
+        "responsibilities": {
+            "description": "Do long-term planning for the Astropy Project as a whole, including:",
+            "details": [
+                "Shape and provide a vision for next major steps in the Project, in consultation with the whole Astropy community.",
+                "Maintain a large-scale view of the project (both from developer and user perspectives).",
+                "Provide points of contact for external stakeholders on the above (with the CoCo).",
+                "Monitor efforts on the topic above to ensure they are being done as imagined.",
+                "Write grant proposals (with the Finance Committee).",
+                "Oversee the Astropy Project Roadmap."
+            ]
```
Cadair, line 73:
```suggestion
                "Engage with, and collect input from, the the whole Astropy Community on project vision and priorities."
            ]
```

eteq:
Alright, this has all been cleaned up, reflecting various feedback both here and the coordination meeting, and should be ready to merge.

ceb8:
Thanks @eteq !

PR #309, opened 2019-03-15
proposed affiliated package: sbpy

mommermi:
This proposal requests affiliated status for `sbpy`, a Python module for small-body planetary astronomy. `sbpy` builds significantly upon astropy functionality and implements well-tested functionality for a variety of use cases in asteroid and comet research. 

This module is still under heavy development and will soon reach v1.0 status, will be available through pypi, and will hopefully obtain a reference with JOSS. The development will be finished by Summer 2021, as this project is funded through a NASA PDART grant. 

A project overview is available at http://sbpy.org
The repository is located at https://github.com/NASA-Planetary-Science/sbpy
The documentation is available at https://sbpy.readthedocs.io/en/latest/

In case there are questions, please don't hesitate to contact me!

PR #587, opened 2024-05-13
Add Thomas Robitaille as fits maintainer

eteq:
Just what it says on the tin!  Note this is a follow-up of #309, as @astrofrog realized there that he was willing to take on this role

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -434,8 +434,8 @@
             {
                 "role": "astropy.io.fits",
                 "people": [
-                    "Thomas Robitaille",
-                    "Simon Conseil"
+                    "Simon Conseil",
+                    "Thomas Robitaille"
                 ]
             },
             {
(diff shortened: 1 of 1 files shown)


roles.json
```diff
@@ -434,6 +434,7 @@
             {
                 "role": "astropy.io.fits",
                 "people": [
+                    "Thomas Robitaille",
```
pllim, line 437: What does this order mean and is Simon okay with it?

mhvk: Indeed, might as well keep order alphabetical as is the case for all the other subpackages.

pllim (approved):
It's been over 2 weeks comment period and no objection. Be careful what you ask for, @astrofrog ! 😸

roles.json
pllim: I fixed it.

pllim:
He is already a core maintainer, no follow-up needed.

PR #588, opened 2024-05-13
remove the warning about master->main

eteq:
It's been 3 years so I think it's safe to do this!


pllim (approved):
This isn't a heavy traffic repo so I am not surprised if someone has really old checkout out there. But then again, GitHub has better instructions anyway, so I think this is okay to remove. Thanks!

PR #591, opened 2024-05-20
Replace devdocs with direct RTD URL

pllim:
Replace devdocs with direct RTD URL for developement docs.

HTTP is not secure and HTTPS for "devdocs" does not work.


eteq (approved):
LGTM!

eteq:
Some backstory for the record: devdocs was originally used because we weren't sure very early on if RTD would be a long-term solution so wanted to use a URL that we could control.  That seems like an unlikely change at this point (over a decade later...), so it's easier to just make the change @pllim did here instead of figure out how to make the redirects work correctly.

PR #475, opened 2022-02-10
Add general maintainers role for astropy core package

hamogu:
This has been discussed in the dev telecon 2022-02-03 and there was general agreement to add such a role. This PR is a draft of that role description; unfortunately I don't remember exactly how we described this role in the telecon, so this is based on my notes, my recollection and the discussion in https://github.com/astropy/astropy-project/issues/136

@bsipocz You had a very good description of this role in the last dev telecon, but I was not fast enough to write that down. Could you please have a look at this and comment or suggest changes to include what you said?

x-ref: https://github.com/astropy/astropy-project/issues/136

PR #614, opened 2024-08-10
Roles: Move general maintainers in same section as sub-package maintainers

pllim:
In https://github.com/astropy/astropy.github.com/pull/475 , there is no explanation why they couldn't be in the same section. Seems weird to have almost duplicate text as different rows. But is it allowed to have "sub-roles" and "responsibilities" not match?

https://output.circle-artifacts.com/output/job/e76af6f1-9bbd-4d0f-bb6f-02e3374ff603/artifacts/0/html/team.html

 xref https://github.com/astropy/astropy.github.com/issues/519

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -336,31 +336,19 @@
         }
     },
     {
-        "role": "Core astropy package maintainer (general)",
-        "url": "Core_package_general_maintainer",
-        "people": [
-            "Derek Homeier",
-            "Pey Lian Lim",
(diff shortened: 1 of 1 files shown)


hamogu (approved):
It's probably just because in #475 I did not understand how the json should be organized. No particular reason why it's not been done this way.

mhvk:
Makes sense!

pllim:
Thanks for the approvals! Merging.

PR #613, opened 2024-08-09
Roles: Consolidate infrastructure roles

pllim:
I finally got around to "Step 1" from the [roles breakout session from Astropy Coordination Meeting 2023](https://docs.google.com/document/d/1sBImbw1p4mM3VYWGQJDRbj2N3uqP5Ed_Wfb8BDjJj2o/edit#heading=h.lap2vjp9ad1z).

Who attended: me, @eteq, @hamogu, @weaverba137, @aaryapatil, @saimn 

Exact notes that is relevant here:

* Expand infrastructure
  * Change Specialist to Project Infrastructure Tech Lead and reduce scope to "set up and maintain CI” and “moving between repo" because others are covered by “core” roles already.
  * Have sub-roles according to broad categories w.r.t. [flow chart from State of Infrastructure this morning](https://docs.google.com/presentation/d/131opPCTy5OWz74G7K88iDn8n3pePSCbjQ5QdHVn7Ckw/edit#slide=id.g23c3c3dc480_0_57).

My interpretation of those notes is this PR.

Out of scope: Adding/removing people.

Preview: https://output.circle-artifacts.com/output/job/c59e5d73-47f0-4fb7-a8d5-03f8163ad418/artifacts/0/html/team.html

PR #615, opened 2024-08-10
Roles: Sub-role in italics instead of bold

pllim:
Weird to have sub-role suddenly be bold font, and thus overshadowing other roles of similar status but just not sub-role.

https://output.circle-artifacts.com/output/job/8e1aeeef-348d-4532-810b-f2420c7a54a8/artifacts/0/html/team.html

xref https://github.com/astropy/astropy.github.com/issues/519

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -125,7 +125,7 @@ $( document ).ready(function(){
                         if (index > 0) {
                             blocks += '<br>';
                         }
-                        blocks += '<strong>' + resp["subrole-head"] + '</strong>';
+                        blocks += '<em>' + resp["subrole-head"] + '</em>';
                     }
                     blocks += '<p>' + resp["description"] + '</p>' +
                               '<ul>' + detail_list + '</ul>';
--- a/css/style.css
+++ b/css/style.css
@@ -24,6 +24,10 @@ time, mark, audio, video {
 	vertical-align: baseline;
 }
 
+em {
+  font-style: italic;
+}
+
 /* HTML5 display-role reset for older browsers */
 article, aside, details, figcaption, figure,
 footer, header, hgroup, menu, nav, section {


js/functions.js
```diff
@@ -125,7 +125,7 @@ $( document ).ready(function(){
                         if (index > 0) {
                             blocks += '<br>';
                         }
-                        blocks += '<strong>' + resp["subrole-head"] + '</strong>';
+                        blocks += '<span style="font-style:italic">' + resp["subrole-head"] + '</span>';
```
mhvk, line 128: Isn't italics just `<em>`?

mhvk:
Result looks good (this is mostly `learn`, right?)

js/functions.js
pllim: I didn't know about `<em>` but `<i>` didn't work. Isn't this more explicit though? According to https://www.w3schools.com/tags/tag_em.asp , "The content inside is *typically* displayed in italic." So it isn't always guaranteed?

pllim:
> (this is mostly learn, right?)

Currently, yes, but it would affect also infrastructure if #613 is affected, and generally any category that splits up into sub-roles.

js/functions.js
mhvk: It doesn't really matter to me, but I like the basic html blocks - let the browser display it in the best way.

pllim: I feel the opposite because sometimes what browser thinks is best is not what I intended. Let's see what others say. Thanks for the feedback!

neutrinoceros: > So it isn't always guaranteed?

Indeed it's not. I would still advise to stick with what works, and just be aware that it may look different depending on user setups... which is always the case (pun intended ?) as soon as you reach for something like fonts (and who doesn't ?).

hamogu: I was tempted not to comment, because it seems like bikeshedding, but then I remembered that I am one of the website maintainers and maybe I should pretend to know something about websites: While it's not in the spec, all browsers I've ever used display `<em>` as italics and it's way simpler and way less likely to clash with something else (say, a browsers "reader view", or - more importantly- a screen reader for visually impaired, which will likely have a mechanism to read `<em>` in some say that is different form normal text, but is unlikely to do anything useful with CSS markup) because it's the most basic HTML.

pllim: OK, changed it to `em` and added Marten as co-author. Thanks for your feedback!

pllim: Actually, when I look at https://output.circle-artifacts.com/output/job/095470ee-dacf-43ea-ba78-c186ad35f1ee/artifacts/0/html/team.html with this change, it does not appear italic to me. Is it just me?

Screenshot 2024-08-12 110620

hamogu: It's because we explicitly reset the meaning of `em` (and a number of other html tags) in this line
```
html, body, div, span, applet, object, iframe,
h1, h2, h3, h4, h5, h6, p, blockquote, pre,
a, abbr, acronym, address, big, cite, code,
del, dfn, em, img, ins, kbd, q, s, samp,
small, strike, strong, tt, var,
b, u, i, center,
dl, dt, dd, ol, ul, li,
fieldset, form, label, legend,
table, caption, tbody, tfoot, thead, tr, th, td,
article, aside, canvas, details, embed,
figure, figcaption, footer, header, hgroup,
menu, nav, output, ruby, section, summary,
time, mark, audio, video {
	margin: 0;
	padding: 0;
	border: 0;
	font-size: 100%;
	font: inherit;
	vertical-align: baseline;
}
```
in our `style.css`.

It's done to reduce browser inconsistencies (none of the browser default matter, because it's all defined explicitly by us), on the other hand, that means that we just get the default we defined that that does not make em look any different from normal text.

I'm really not sure what to do, here. I mean, I do know what do to, but I don't want this simple PR to balloon into a redesign of our web styles. We could remove the `em` in the line I gave above, thus getting browser defaults again. Or, we could add the following lines to out `style.css`:
```
em {
  font-style: italic;
}
```
which should (not tested) set the font style to italic (making it guaranteed that em is italic on our page for every browser). 
But, do we use `<em>` tags in other places and would change that? Do we want to review the rest of our website for that? Is it really worth the effort? Is it even worth the effort to write this long a reply?

OK, here's my suggestion: Add 
```
em {
  font-style: italic;
}
```
to style.css. If we used `em` in other places and it did not actually emphasize anything, it's abut that we are fixing now.

pllim: Thanks! Hopefully I added it correctly. Also added you as co-author.

Also looks like the sub-role is the only thing that is using `<em>`.

hamogu (approved):

PR #621, opened 2024-09-20
Feature Astropy Project awards more prominently

taldcroft:
We should highlight the prestigious awards of the project on the main landing page instead of burying them deep on the team page.

In addition, let's shout about the AAS Award on the front page! We can plan to remove this banner in two or three months.

### DIFF
--- a/index.html
+++ b/index.html
@@ -161,7 +161,19 @@ window.onload = function() {
 		<a class="button" href="https://numfocus.org/donate-to-astropy" target="_blank">Donate to Astropy</a>
 	</section>
 
-	<section class="whatsnext">
+    <section>
+        <h1 id="awards">Awards<a class="paralink" href="#awards" title="Permalink to this headline">¶</a></h2>
+        <p>
+          <ul>
+            <li/><a href="https://aas.org/press/astropy-collaboration-receive-2025-berkeley-prize">Lancelot M. Berkeley–New York Community Trust Prize for Meritorious Work in Astronomy</a> (2025)
+            <li/><a href="https://accreditations.ioppublishing.org/9ce7a38e-8ef1-4aa3-86f4-5760022a1574#gs.0lk076">IOP Publishing Top Cited Paper Awards North America</a> (2023)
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -115,6 +115,10 @@
 
 </section>
 
+<section class="whatsnext">
+    <h1 id="news">Congratulations</h1> The Astropy Project has been awarded the <a href="https://aas.org/press/astropy-collaboration-receive-2025-berkeley-prize">Lancelot M. Berkeley–New York Community Trust Prize for Meritorious Work in Astronomy</a> for 2025!
```
pllim, line 119: "Congratulations" seems a little weird. Are we congratulating ourselves? Maybe "latest news" but then we have the extra burden to make sure the section actually have latest news. 🤷‍♀️ 

When we sort out https://github.com/astropy/astropy-project/issues/431 , we can add that here too, maybe.

pllim: Also maybe add a link to jump to Awards below?

taldcroft: This is congratulating all the contributors and community members (and funding sources) that made it possible!

taldcroft: > Also maybe add a link to jump to Awards below?

I think it's OK without another link (less is more, just focusing on the link that matters here), and this section will go away in a few months anyway.

taldcroft:
> Hmm Astropy is on the actual award homepage now but less details. Should we switch over or not?
> 
> https://aas.org/grants-and-prizes/lancelot-m-berkeley-new-york-community-trust-prize-meritorious-work-astronomy

I think we should stick with the full press release in both places since it provides so much more information. I think the AAS is missing an opportunity by not having the Prize summary page not link to the individual citations for each year, but that isn't our issue to fix.

index.html
pllim: How many months are you thinking of? Need a number for the follow-up issue. Six months?

pllim: Oh, sorry. I just saw you mentioned 2-3 months in the original post. That sounds good.

pllim (approved):
LGTM. Thanks!

PR #633, opened 2024-10-14
Remove duplicate Ed Slavich

pllim:
Remove duplicate Ed Slavich because it is the same as Edward Slavich.

Pinging @astropy/astropy-project-release-team because I think this list is updated during releases?


pllim:
https://github.com/astropy/astropy/pull/17322 is merged, so I will merge this too to be consistent.

PR #648, opened 2025-01-29
Update index.html per Executive Order

taldcroft:
Per the memo from Karla Smith Jackson "Initial Guidance Regarding Diversity, Equity, Inclusion, Accessibility (DEIA) Executive Orders", this changes the community code of conduct link reference on the astropy.org landing page. Astropy has two NASA grants which fall under the jurisdiction of the order based on other email guidance.

### DIFF
--- a/index.html
+++ b/index.html
@@ -104,8 +104,8 @@ window.onload = function() {
 		<p>The Astropy Project is a community effort to develop a <a href="https://docs.astropy.org">common core
 		package</a> for Astronomy in Python and foster an ecosystem of <a href="affiliated/index.html">interoperable
 		astronomy packages</a>.</p>
-        <p>The Astropy community <a href="https://www.astropy.org/code_of_conduct.html">is committed to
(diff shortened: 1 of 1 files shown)


hamogu:
I absolutely disagree with the notion that an American present can control what we do and say in Astropy, which is largely volunteer-run and a worldwide organization. If we want DEI, we can do that - we just can't to it with NASA funding, i.e. it's something only volunteers (preferred not from the US) can work on.
(Though NASA has the full authority to revoke that Astropy grants if they deem us to be a "woke" organization, I think we need a discussion if that thread forces us to adapt out values, or if we take the risk to loose the NSASA funding and stand by our values. And that discussion should happen in a broader forum, not just in the PR to the website.)

However, in this particular case, I don't see what the link was not called "code of conduct" before, because that's what it is linking to; we I don't think we need that discussion fir this particular PR.

hamogu (approved):

Issue #631, opened 2024-10-10
Remove "Congratulations to all Astropy contributors" after 2025-01-17


pllim:
This is a follow-up of https://github.com/astropy/astropy.github.com/pull/621 where @taldcroft said the top "Congratulations to all Astropy contributors"  can be removed after "2 or 3 months". It was merged on 2024-09-23.

Also remove the banner added in #636 (if it is merged). That is, empty the content but do not remove the file.

eteq:
Now that I think about it, how about we leave this in until after the AAS talk?  Since that's specifically what the prize is about.  So it can go away 2025-1-17.

PR #649, opened 2025-02-05
Remove congratulations banners

pllim:
Remove congratulations banners about AAS award and 10k citations because it had been a few months.

Fix https://github.com/astropy/astropy.github.com/issues/631


mhvk (approved):
Since the banner had started to irritate me after having seen it too many times, yes, please! (Any next time I'd suggest not to do it on any of the inner pages, though perhaps that excludes too many people who get there by googling...)

pllim:
It is unlikely we will hit 20k citations any time soon, right? Plenty of time to ponder lol

PR #653, opened 2025-02-28
Add banner for survey announcement

jeffjennings:
This PR adds a banner (like the recent one for 10k citations to Astropy paper I) that advertises the astropy user survey, which will close on April 1. I should've thought of adding this banner sooner, but I'm hoping to get this merged quickly, as the survey just opened. Thanks!

### DIFF
--- a/index.html
+++ b/index.html
@@ -115,6 +115,11 @@ window.onload = function() {
 
 </section>
 
+<section class="whatsnext">
+    	<h1 id="news">Astropy User Survey open until April 1</h1> 
+    The Astropy Project is running its first <a href="https://forms.gle/KX2cNuFsQYgVhF178">user survey</a>. Contribute your input by April 1!
+</section>
+	
 	<section class="install">
 		<h1 id="install-astropy">Install Astropy<a class="paralink" href="#install-astropy" title="Permalink to this headline">¶</a></h1>
         There are a number of ways of installing the latest release of the astropy core package. If you normally use <code>pip</code> to install Python packages, you can do:


jeffjennings:
@pllim Would you mind reviewing this? I saw that you were adding/removing the 10k citations announcement content, which I just copied/modified here. Thank you!

adrn:
I would prefer to ask the webpage maintainers on this rather than singling out @mhvk as one potentially negative opinion! I'm sure there will be a diversity of opinions but I don't think this is something we need to ask the entire project about...

adrn:
Can you request reviews from @dhomeier and @eteq?

mhvk:
FWIW, I liked the 10000 citations banner but felt that after having been there for a while, it should no longer be on *every* page, just on the landing page.

For the user survey, having it everywhere makes sense, let's just not have it there for too long (or can we ensure someone doesn't see it more than once a week or so?? but that probably needs cookies, etc., so not great...). 

Anyway, bottom line: let's be sure that there is an expiration day set now!

jeffjennings:
> FWIW, I liked the 10000 citations banner but felt that after having been there for a while, it should no longer be on _every_ page, just on the landing page.
> 
> For the user survey, having it everywhere makes sense, let's just not have it there for too long (or can we ensure someone doesn't see it more than once a week or so?? but that probably needs cookies, etc., so not great...).
> 
> Anyway, bottom line: let's be sure that there is an expiration day set now!

Thanks Marten! That sounds good - we're keeping the survey open until April 1, so I'm happy to submit another PR to remove the banner on April 2.

jeffjennings:
Would someone mind merging this in? I know the website maintainers are tagged for review, but anyway I don't think this is controversial (and it's temporary). The more people visiting the site we can reach, the bigger impact it can have on the user survey!

dhomeier (approved):
Thanks, I haven't seen any adverse opinions in the end!

jeffjennings:
Thank you @dhomeier and @hamogu! I'll submit another PR in 1 month to remove the banner.

PR #550, opened 2023-09-15
Proposed affiliated package: PetroFit (pyOpenSci/software-submission#159)

robelgeda:
EDIT: Superseded by https://github.com/pyOpenSci/software-submission/issues/159

PetroFit is a Python package designed for calculating Petrosian properties, such as radii and concentration indices, as well as fitting galaxy light profiles. In particular, PetroFit includes tools for performing accurate photometry, segmentations, Petrosian profiling, and Sérsic fitting. PetroFit is based on PhotUtils and AstroPy; and we hope to contribute the more generalized aspects of our code upstream for other packages make use of. For more information please visit the PetroFit [repo](https://github.com/PetroFit/petrofit/tree/main) or [documentation](https://petrofit.readthedocs.io/en/latest/).

cc @crawfordsm

PR #656, opened 2025-03-04
pyOpenSci affiliated package listing: Use labels instead of partners field

pllim:
As suggested in https://github.com/astropy/astropy.github.com/pull/550#issuecomment-2695927891 . Also see:

* https://github.com/pyOpenSci/pyopensci.github.io/pull/576


hamogu (approved):
Can't say I understand javascript, but it makes sense to me and seems to work.

PR #658, opened 2025-03-04
Update google analytics measurement ID

jeffjennings:
For the current website analytics, Google analytics suggests updating the old measurement ID to a new format because "Legacy UA tags with connected site tags could lead to inconsistent data reports". This PR updates the measurement ID from "UA-*" to the new identifier. I've verified that it is not an issue for the measurement ID to be public, as it has been.

### DIFF
--- a/js/analytics.js
+++ b/js/analytics.js
@@ -3,7 +3,7 @@
 m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
 })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
-ga('create', 'UA-30968842-1', 'auto');
+ga('create', 'G-R0510VK4B6', 'auto');
 ga('require', 'displayfeatures');
 ga('require', 'linkid', 'linkid.js');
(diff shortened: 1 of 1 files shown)


pllim:
How do we know it works? Do we have to YOLO this one? Anything to be done on RTD also?

pllim:
Also does this close https://github.com/astropy/astropy/issues/17778 ? Thanks!

jeffjennings:
Thanks for the quick feedback! 

The google analytics are already working on all pages of astropy.org, so the current measurement ID is functional, but I thought it best to update it to the newer one, given the warning I quoted (and to be consistent with the newer ID used in https://github.com/astropy/astropy/pull/17845). Both IDs are tied to the same analytics, so it is maybe a bit of YOLO, but I don't see why it wouldn't work. 

There shouldn't need to be any changes on the readthedocs side, as things are already working now.  

And this doesn't close https://github.com/astropy/astropy/issues/17778, but https://github.com/astropy/astropy/pull/17845 does.

pllim (approved):
If nothing else, it will be consistent with core changes. Let's get this in. Thanks!

PR #661, opened 2025-04-04
Remove defunct example gallery from Learn section

pllim:
Example gallery in core lib never really took off, so it was removed. The link will be broken when astropy v7.1 is released.

* astropy/astropy#17270 
* astropy/astropy#16194


ceb8 (approved):
LGTM

pllim:
Thanks for the review!

PR #667, opened 2025-06-17
update what's new entry to 7.1

eteq:
It looks like this never got updated at 7.1 release.  Updating it now!


pllim (approved):
@astropy/astropy-project-release-team , anything we need to do to prevent such oversight in future release?

Thanks!

PR #675, opened 2025-06-27
Remove packages that went through PyOpenSci from our old listing

hamogu:
Since the packges are now listed through PyOpenSci, they appear twice on the Astropy affiliated website once under the heading "Affiliated Packages Registry" and once under the heading "Affiliated Packages Registry (Pre-APE 22)".

We're trying to make that list the most useful for people to find packages and we think it's clearer if each package is listed only once. Thus, this PR removes the older (pre-APE 22) listing towards the bottom on the page.

Both affected packages have been contacted and agree:

https://github.com/StingraySoftware/stingray/issues/926
https://github.com/Cosmoglobe/zodipy/issues/50

### DIFF
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -675,7 +675,6 @@
                 "last-updated": "2017-11-08"
             }
         },
-
         {
             "name": "Halotools",
             "maintainer": "Andrew Hearin, Nick van Alfen <ahearin@nal.gov>",
--- a/affiliated/registry.json
+++ b/affiliated/registry.json
@@ -1,24 +1,5 @@
 {
     "packages": [
-        {
-            "name": "ZodiPy",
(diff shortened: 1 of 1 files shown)


dhomeier (approved):
Formatting issue aside this should be absolutely uncontroversial; I had thought reviewing under pyOpenSci again was already understood as _replacing_ the old Affiliated entry.

pllim (changes requested):
I think we now need to update the text on https://www.astropy.org/affiliated/#affiliated-package-list-pre-ape22 because this sentence is no longer accurate:

"This section is frozen as of March 6, 2024."

The new text should say that we remove dead and duplicated packages.

pllim (approved):
Nvm I just saw https://github.com/astropy/astropy.github.com/pull/677 . Thanks!

PR #679, opened 2025-07-05
Split out a new section Archived/emeritus for affiliated packages

hamogu:
The original intent (at least in my mind) for affiliated packages was to provide the community with a list of high-quality, actively maintained packages for their needs in astronomy. After many years, we have a large number of affiliated packages (good!) and some of them are no longer active, e.g. they have been archived in github by their maintainers. Yet, some of these packages are still used widely and perform well, at least if installed in environments with older Python versions. Thus, we don't want to remove them from the list, but we also want to distinguish them in some way.
    
Here, we split them out into a new section.
    
 This was discussed at the 2025 Astropy Coordination meeting, but is a variant of of the ideas mentioned in the discussion.

### DIFF
--- a/affiliated/index.html
+++ b/affiliated/index.html
@@ -265,6 +265,10 @@ but for now just do a generic link to their listing. -->
 			However, some of those packages still provide unique
 			capabilities that no newer package matches or are still widely used.
 		</p>
+		<p>If there is new active development, the package can be moved back into <a href="#affiliated-package-list-pre-ape22">the pre-APE 22 list</a>;
+			if you are interested in taking over development from one of these packages we recommend contacting the maintainer or reach out to the Astropy
+			community.
+		</p>
 
 		<p>Total number of archived/emeritus affiliated packages: <strong id="total-emeritus-pkgs"></strong></p>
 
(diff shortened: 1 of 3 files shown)


hamogu (commented):
This needs comminity discussion, but I encourage @dhomeier to take a look first before we send it to the mailing list.

hamogu:
We'll have to see what pyOpenSci does, will they add a "deprecated" tag to their listing that we can pick up? Or will they just be removed? In that case, we would add them manually to our section.

But honestly, I think we are getting ahead here. We've had the "affiliated" packages for over decode and only now this comes up. We don't know the state of astropy or the Python packaging ecosystem a decade from now, so I think we should't worry about that too much right now. We need a few cases where that actually happens to see what will work - and those cases are likely years away.

pllim:
Does this mean we should revert this PR

* https://github.com/astropy/astropy.github.com/pull/676

and add SABA to "archived/emeritus" instead?

hamogu:
Good idea! But since it's in a different repro, let's discuss here first and if we agree that this is the way to go, we can make corresponding updates in `astropy-project`.

pllim:
I trust the current Editors to have sound decisions in this matter, so it is up to you and @dhomeier . Thanks! :)

PR #685, opened 2025-08-26
minor name change for spoc

eteq:
update to strategic planning committee member listing to reflect new name for committee (SPOC)

cc @pllim


pllim (approved):
😆

PR #691, opened 2025-10-13
Add APE Editors

pllim:
Establish responsibilities for the new [ape-editor-team](https://github.com/orgs/astropy/teams/ape-editor-team) as introduced by:

* https://github.com/astropy/astropy-APEs/pull/115

This should only be taken out of draft after https://github.com/astropy/astropy-APEs/pull/115 is merged.

It would be initially staffed by CoCo, though ideally it should be eventually handed off to other community members who volunteer following the process established for other roles.

### DIFF
--- a/roles.json
+++ b/roles.json
@@ -276,10 +276,10 @@
         "url": "APE_editor",
         "people": [
             "Clara Brasseur",
-            "Derek Homeier",
-            "Pey Lian Lim",
             "Aarya Patil",
-            "Erik Tollerud"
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Should we add the newly elected CoCo members before we merge this ?

pllim:
Good point. I wonder if we should just swap out CoCo people while we're at it. In that case, we can wait till after handover meeting (to be scheduled).

pllim:
Merged. Thanks!

PR #702, opened 2026-03-16
Adding link to AI policy

ceb8:
Added a link to the AI policy on the contribution page.


contribute.html
```diff
@@ -151,7 +151,7 @@ <h2 id="contribute-code-or-docs">Contribute code or documentation<a class="paral
...
       <li><a href="https://github.com/search?p=2&q=label%3Apackage-novice+label%3Aeffort-high&ref=searchresults&state=open&type=Issues&utf8=%E2%9C%93" class="github-label-effort">Effort-high</a>: longer term issues
       </ul>
-      <p>You may also want to familiarize yourself with the <a href="https://docs.astropy.org/en/latest/#developer-documentation" target="_blank">developer documentation</a>, particularly the <a href="https://docs.astropy.org/en/latest/development/codeguide.html">coding</a> and <a href="https://docs.astropy.org/en/latest/development/docguide.html">documentation</a> guidelines.</p>
+      <p>You may also want to familiarize yourself with the <a href="https://docs.astropy.org/en/latest/#developer-documentation" target="_blank">developer documentation</a>, particularly the <a href="https://docs.astropy.org/en/latest/development/codeguide.html">coding</a> and <a href="https://docs.astropy.org/en/latest/development/docguide.html">documentation</a> guidelines, and our <a href="https://github.com/astropy/astropy-project/blob/main/policies/ai-policy.md">AI usage</a> policy.</p>
```
pllim, line 154: Is this obvious enough? I could almost not find it without looking at the diff to figure out where it was added: https://output.circle-artifacts.com/output/job/1f310ea5-2de2-445a-a2a3-dda383604e02/artifacts/0/html/contribute.html

ceb8: I could put it on its own line?

pllim: @Cadair @hamogu etc, what do you think?

hamogu (approved):
I suggest we leave it there for now and can always move it or make it more prominent if it later turns out to be a problem.
In my view it's more likely that someone obviously violates that policy (and those people are less likely to look in the first case) and we want to on the website so we can point to it when we reject a PR. In that case, it's mostly us who need to know where it is so that we can put an URL somewhere.

People who contribute in good faith will essentially follow this policy automatically anyway.

PR #717, opened 2026-06-17
Remove obsolete getteam.py

astrofrog:
The script updated the team.html file which no longer exists. The credits.rst file is now downloaded directly by conf.py and included into team.rst at Sphinx build time.


pllim (approved):
Was this removed during Stuart's huge refactoring? Thanks for the clean-up.

PR #721, opened 2026-06-29
Remove now suspicious link from Travis CI

pllim:
The link chronically fails linkcheck now with "415 Client Error: Unsupported Media Type for url" and when I manually check the link, at one point, my antivirus blocked a malicious thingy or so it claimed. I don't think this URL is worth keeping.


pllim:
My PRs seem to not get reviewed over here and this fixes the CI, so I am just gonna merge without review.

neutrinoceros:
for reference, you can feel free to ping me in the future if you really want someone else to look over. At least in this instance, I think it's perfectly okay that you merged without a review though. Thanks !

PR #732, opened 2026-09-08
Ignore numfocus shop link blocking linkcheck

pllim:
`403 Client Error: Forbidden for url`


pllim:
Zenodo site is temporarily down or something. Linkcheck failure is unrelated to numfocus now.

PR #734, opened 2026-09-11
Fix RAW data URL for pyOpenSci crosslisting

### LINKED ISSUE #733: Affiliated registry cross-listing is broken (again)
https://www.astropy.org/affiliated/index.html#affiliated-packages-registry should have same listing as https://www.pyopensci.org/packages/astropy/ but it is not showing.

pllim:
Fix https://github.com/astropy/astropy.github.com/issues/733

### DIFF
--- a/affiliated/index.rst
+++ b/affiliated/index.rst
@@ -228,7 +228,7 @@ It is important that you understand your reviews will be done in public. There i
     </script>
 
     <script>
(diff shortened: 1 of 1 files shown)


pllim:
OK I think this fixes it. Since the problem is embarrassing, I am just gonna merge without review.

